PR #7, opened 2016-09-14
Remove async prefix

gregw:
https://github.com/GoogleCloudPlatform/jetty-runtime/issues/20
parameterized docker image tags and labels for ease of CI

### DIFF
--- a/openjdk8/pom.xml
+++ b/openjdk8/pom.xml
@@ -67,7 +67,7 @@
             <configuration>
               <imageName>openjdk8</imageName>
 	      <imageTags>
-	        <imageTag>${docker.label.prefix}8-jre</imageTag>
+	        <imageTag>${docker.tag}</imageTag>
 	      </imageTags>
               <dockerDirectory>${project.build.directory}/docker</dockerDirectory>
             </configuration>
@@ -81,7 +81,7 @@
 	    <configuration>
 	      <imageName>openjdk8</imageName>
 	      <imageTags>
-	        <imageTag>${docker.label.prefix}8-jre</imageTag>
+	        <imageTag>${docker.tag}</imageTag>
 	      </imageTags>
 	    </configuration>
 	  </execution>
--- a/openjdk8/src/main/docker/Dockerfile
+++ b/openjdk8/src/main/docker/Dockerfile
@@ -12,12 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM gcr.io/google_appengine/base
+FROM ${docker.appengine.image}
 ENV DEBIAN_FRONTEND noninteractive
 
 # Create env vars to identify image
 ENV GAE_IMAGE_NAME ${project.artifactId}
-ENV GAE_IMAGE_LABEL ${docker.label.prefix}8-jre
+ENV GAE_IMAGE_LABEL ${docker.tag}
 ENV GAE_IMAGE_VERSION ${project.version}
 
 # Update debian 
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,8 @@
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <maven.build.timestamp.format>yyyyMMddHHmm</maven.build.timestamp.format>
     <appengine.api.version>1.9.40</appengine.api.version>
-    <docker.label.prefix>async-</docker.label.prefix>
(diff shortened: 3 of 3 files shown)


openjdk8/src/main/docker/Dockerfile
```diff
@@ -12,12 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM gcr.io/google_appengine/base
+FROM ${docker.appengine.image}
```
meltsufin: What's the motivation for externalizing this into a property? If it's only used here, I'd rather avoid the extra level of indirection.

pom.xml
```diff
@@ -34,7 +34,8 @@
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <maven.build.timestamp.format>yyyyMMddHHmm</maven.build.timestamp.format>
     <appengine.api.version>1.9.40</appengine.api.version>
-    <docker.label.prefix>async-</docker.label.prefix>
+    <docker.tag>8-jre</docker.tag>
```
meltsufin: I think the default tag should be very specific and include the build timestamp.

openjdk8/pom.xml
```diff
@@ -81,7 +81,7 @@
 	    <configuration>
 	      <imageName>openjdk8</imageName>
```
meltsufin: we should rename this to "openjdk".

openjdk8/src/main/docker/Dockerfile
gregw: For CI we will want to build openjdk-runtime, jetty-runtime and jetty-gae-compat runtime in a series, but we may also have other CI builds of branches and alternatives, thus their is the risk of collision in the docker namespace.   Also eventually the default values for the jetty-runtime and jetty-gae-compat will be to pull released images from gcr.io.

So I parametrized, so properties can be used to control the docker name space and link up the images in CI build chains, without needing to release into gcr.io etc.

I guess it is not really needed at this bottom layer... but this keeps the pattern and does allow for other base images to be easily injected.

openjdk8/pom.xml
gregw: ok

pom.xml
gregw: I build ID will be easier than a timestamp.  Timestamps will make CI build chains difficult or at least fragile if they are datestamps rather than timestamps.

openjdk8/src/main/docker/Dockerfile
meltsufin: OK, makes sense.

pom.xml
meltsufin: I see your point about potential difficulty in working with timestamps as versions. So, we should discuss possible workarounds or push back on the whole idea. Maybe we should use something like git commit SHA instead.

gregw: Can we leave the issue of build ids / timestamps to a different Issue/PR, as it may be a little complex, or at least need careful though with how it is going to interact with version and CI.

openjdk8/pom.xml
gregw: Actually,  it is currently openjdk8 to match the module name and directory name, in anticipation of there soon being an openjdk9

I thought the decision was to keep the version in the name when parallel versions were likely to be maintained concurrently - and thus avoid lots of branches.

I guess we could have directory/module openjdk8 builds openjdk:8-buildID rather than openjdk8:buildID ?

meltsufin: Yes,I don't think the image name necessarily has to match the directory structure of the code. The reason we wanted `openjdk` is to be able to provide a `openjdk:latest` tag for users who just want to be on the latest version of Java.

gregw: OK - but can we bump that rename to a different issue/PR as it probably comes into the version/buildID considerations.  

We may even need multiple names/tags for the one image as many users will want to get the latest java8 with `openjdk8:latest`

So for now can we strip out the unneeded async- prefix?

pom.xml
meltsufin: Sure. Here's the issue: https://github.com/GoogleCloudPlatform/openjdk-runtime/issues/8

openjdk8/pom.xml
meltsufin: Yep, https://github.com/GoogleCloudPlatform/openjdk-runtime/issues/9.

meltsufin:
LGTM

Issue #9, opened 2016-09-14
Rename image from openjdk8 to openjdk


meltsufin:

meltsufin:
Done in #15.

Issue #8, opened 2016-09-14
Default docker image version label should contain a unique build id


meltsufin:

joakime:
commit id hashes are not unique enough.
the same commit id will be in many branches and pull requests.

you can still have collisions in the docker image tags and deployed uris (for testing)

meltsufin:
@joakime What makes you say that the git commit hash is not unique enough? If you build by checking out a specific commit, wouldn't it always build from the exact same source?

PR #10, opened 2016-09-15
Issues #8 and #9 improve docker image names and tags

gregw:
This is a speculative PR for #8 and #9 to start continue the discussion of how we want the
image names and tags to be.

This now builds an image with the following tags (currently with SNAPSHOT):
 - openjdk:8
 - openjdk:8u102
 - openjdk:8u102-1-SNAPSHOT
 - openjdk:8u102-1-SNAPSHOT-2016-09-15_01_31
 - openjdk:8u102-2016-09-15_01_31
 - openjdk:latest

This allows a dependent image to use:

 - the latest version of openjdk
 - the latest version of openjdk:8
 - the latest build of a specific version of openjdk:8u102
 - a specific build of a specific version of openjdk:8u102-1-SNAPSHOT by version
 - a specific build of a specific version of openjdk:8u102-2016-09-15_01_31 by timestamp
 - a specific build of a specific version of openjdk:8u102-1-SNAPSHOT-2016-09-15_01_31 by version and timestamp

The version is now just a single digit release number and if future we could expect multiple
versions of java to be built under a single release number:

 - openjdk:8u108-42-2017-02-01_12_00
 - openjdk:8u108-42
 - openjdk:8u108
 - openjdk:8
 - openjdk:9u2-42-2017-02-01_12_01
 - openjdk:9u2-42
 - openjdk:9u2
 - openjdk:9

It may be a bit over the top having both release number and timestamp, but tags are cheap!

### DIFF
--- a/logging/pom.xml
+++ b/logging/pom.xml
@@ -21,7 +21,7 @@
   <parent>
     <groupId>com.google.cloud.runtimes</groupId>
     <artifactId>openjdk-parent</artifactId>
-    <version>0.1.0-SNAPSHOT</version>
+    <version>1-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
   <name>Google App Engine Logging Support</name>
--- a/openjdk8/pom.xml
+++ b/openjdk8/pom.xml
@@ -67,8 +67,6 @@
             <configuration>
               <imageName>openjdk</imageName>
 	      <imageTags>
-	        <imageTag>${docker.tag.prefix}${openjdk8.version}-${project.version}-${maven.build.timestamp}</imageTag>
-	        <imageTag>${docker.tag.prefix}${openjdk8.version}-${project.version}</imageTag>
 	        <imageTag>${docker.tag.prefix}${openjdk8.version}-${maven.build.timestamp}</imageTag>
 	        <imageTag>${docker.tag.prefix}${openjdk8.version}</imageTag>
 	        <imageTag>${docker.tag.prefix}8</imageTag>
@@ -85,8 +83,6 @@
 	    <configuration>
 	      <imageName>openjdk</imageName>
 	      <imageTags>
-	        <imageTag>${docker.tag.prefix}${openjdk8.version}-${project.version}-${maven.build.timestamp}</imageTag>
-	        <imageTag>${docker.tag.prefix}${openjdk8.version}-${project.version}</imageTag>
 	        <imageTag>${docker.tag.prefix}${openjdk8.version}-${maven.build.timestamp}</imageTag>
 	        <imageTag>${docker.tag.prefix}${openjdk8.version}</imageTag>
 	        <imageTag>${docker.tag.prefix}8</imageTag>
--- a/openjdk8/pom.xml
+++ b/openjdk8/pom.xml
@@ -67,11 +67,11 @@
             <configuration>
               <imageName>openjdk</imageName>
 	      <imageTags>
-	        <imageTag>${openjdk8.version}-${project.version}-${maven.build.timestamp}</imageTag>
-	        <imageTag>${openjdk8.version}-${project.version}</imageTag>
-	        <imageTag>${openjdk8.version}-${maven.build.timestamp}</imageTag>
-	        <imageTag>${openjdk8.version}</imageTag>
-	        <imageTag>8</imageTag>
+	        <imageTag>${docker.tag.prefix}${openjdk8.version}-${project.version}-${maven.build.timestamp}</imageTag>
+	        <imageTag>${docker.tag.prefix}${openjdk8.version}-${project.version}</imageTag>
+	        <imageTag>${docker.tag.prefix}${openjdk8.version}-${maven.build.timestamp}</imageTag>
+	        <imageTag>${docker.tag.prefix}${openjdk8.version}</imageTag>
+	        <imageTag>${docker.tag.prefix}8</imageTag>
 	      </imageTags>
               <dockerDirectory>${project.build.directory}/docker</dockerDirectory>
             </configuration>
@@ -85,11 +85,11 @@
 	    <configuration>
 	      <imageName>openjdk</imageName>
 	      <imageTags>
-	        <imageTag>${openjdk8.version}-${project.version}-${maven.build.timestamp}</imageTag>
-	        <imageTag>${openjdk8.version}-${project.version}</imageTag>
-	        <imageTag>${openjdk8.version}-${maven.build.timestamp}</imageTag>
-	        <imageTag>${openjdk8.version}</imageTag>
-	        <imageTag>8</imageTag>
+	        <imageTag>${docker.tag.prefix}${openjdk8.version}-${project.version}-${maven.build.timestamp}</imageTag>
+	        <imageTag>${docker.tag.prefix}${openjdk8.version}-${project.version}</imageTag>
+	        <imageTag>${docker.tag.prefix}${openjdk8.version}-${maven.build.timestamp}</imageTag>
+	        <imageTag>${docker.tag.prefix}${openjdk8.version}</imageTag>
+	        <imageTag>${docker.tag.prefix}8</imageTag>
 	      </imageTags>
 	    </configuration>
 	  </execution>
--- a/openjdk8/pom.xml
+++ b/openjdk8/pom.xml
@@ -21,7 +21,7 @@
   <parent>
     <groupId>com.google.cloud.runtimes</groupId>
     <artifactId>openjdk-parent</artifactId>
-    <version>0.1.0-SNAPSHOT</version>
+    <version>1-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
   <name>Google App Engine Image for OpenJDK (openjdk8)</name>
@@ -65,9 +65,13 @@
               <goal>build</goal>
             </goals>
             <configuration>
-              <imageName>openjdk8</imageName>
+              <imageName>openjdk</imageName>
 	      <imageTags>
-	        <imageTag>${docker.tag}</imageTag>
+	        <imageTag>${openjdk8.version}-${project.version}-${maven.build.timestamp}</imageTag>
+	        <imageTag>${openjdk8.version}-${project.version}</imageTag>
+	        <imageTag>${openjdk8.version}-${maven.build.timestamp}</imageTag>
+	        <imageTag>${openjdk8.version}</imageTag>
+	        <imageTag>8</imageTag>
 	      </imageTags>
               <dockerDirectory>${project.build.directory}/docker</dockerDirectory>
             </configuration>
@@ -79,9 +83,13 @@
 	      <goal>removeImage</goal>
 	    </goals>
 	    <configuration>
-	      <imageName>openjdk8</imageName>
+	      <imageName>openjdk</imageName>
 	      <imageTags>
-	        <imageTag>${docker.tag}</imageTag>
+	        <imageTag>${openjdk8.version}-${project.version}-${maven.build.timestamp}</imageTag>
+	        <imageTag>${openjdk8.version}-${project.version}</imageTag>
+	        <imageTag>${openjdk8.version}-${maven.build.timestamp}</imageTag>
+	        <imageTag>${openjdk8.version}</imageTag>
+	        <imageTag>8</imageTag>
 	      </imageTags>
 	    </configuration>
 	  </execution>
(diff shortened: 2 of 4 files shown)


gregw:
@meltsufin thoughts?

joakime:
I would have expected to see `openjdk:{branch}` as some sort of option too.

This will pay benefits when we get into CI with PR builds, as it should also make `openjdk:PR-#` images (not stepping on the toes of the normal builds)

meltsufin:
I like that Maven has the `maven.build.timestamp` property that we can use. I was thinking that by default we should only tag with a fully qualified tag, potentially including the branch, as Joakim suggests -- something like `openjdk:{jdk_version}-{branch}-{project_version}-{build_timestamp}`. Adding additional tags or retagging would be part of the release process.

I'm still unsure of including both project version and build timestamp in the version tag. We should probably standardize on one or the other, or something different like the git commit hash.

aslo:
@meltsufin I agree that we should default to fully qualified tags, though I'm not sure about the format you proposed.

I think `openjdk:{jdk_version}-{project_version}-{build_timestamp}` would be fine. Since commits can appear on more than one branch, I don't think `branch` really belongs as part of the format. I believe @joakime was just suggesting using `openjdk:{branch}` as a full tag, which would be useful in a CI environment.

joakime:
A docker tag is sufficient, yes.

Just have to be careful to keep the CI systems builds separate enough to not overlap between the docker build and docker tag steps.

Eg (example timeline):

  - build master/4 - docker build openjdk:latest
  - build issue-8/1 - docker build openjdk:latest
  - build master/4 - docker tag openjdk:latest openjdk:master
  - build issue-8/1 - docker tag openjdk:latest openjdk:issue-8

In this case, the `openjdk:master` will not be what is expected if we allow parallel builds to improperly to share a common local docker images repository.

This is more of an issue with a proper CI setup.  

But it helps to have the `openjdk:{branch}` tag creation be part of the build.
Suggestion:  do this as a automatic maven build profile.  if property ... uhm ... `DOCKER_BRANCH_TAG` exists, use it to create a docker tag.

gregw:
I've added ${docker.tag.prefix}, which is currently set to empty string, but you can build with:
```
mvn -Ddocker.tag.prefix=test- install
```
and add a prefix to all tags generated:
```
[2043] docker images
REPOSITORY                          TAG                                      IMAGE ID            CREATED             SIZE
openjdk                             latest                                   9ab10aacdfc3        29 seconds ago      549.1 MB
openjdk                             test-8                                   9ab10aacdfc3        29 seconds ago      549.1 MB
openjdk                             test-8u102                               9ab10aacdfc3        29 seconds ago      549.1 MB
openjdk                             test-8u102-1-SNAPSHOT                    9ab10aacdfc3        29 seconds ago      549.1 MB
openjdk                             test-8u102-1-SNAPSHOT-2016-09-15_21_35   9ab10aacdfc3        29 seconds ago      549.1 MB
openjdk                             test-8u102-2016-09-15_21_35              9ab10aacdfc3        29 seconds ago      549.1 MB
```

gregw:
Oh0 I just noticed that @meltsufin suggested the branch tag part to be after the java version... let me swap this...

gregw:
Note that I've got a [branch of jetty-runtime](https://github.com/jetty-project/gcloud-jetty-runtime/tree/master-openjdk-8-9) ready with the matching changes that results in images like:
```
jetty:latest
jetty:9
jetty:9.3.8.v20160314  
jetty:9.3.8.v20160314-1-SNAPSHOT 
jetty:9.3.8.v20160314-1-SNAPSHOT-2016-09-16_04_33  
jetty:9.3.8.v20160314-2016-09-16_04_33  
```

But I'll hold off making the PR until we decide if `version-build-timestamp` is too much or too little.

If I was to drop any of these, it would be the `version-timestamp`   So we would be left with releases looking like:
```
openjdk:             latest
openjdk:             8
openjdk:             8u102
openjdk:             8u102-1
openjdk:             8u102-1-2016-09-15_21_35
jetty:                   latest
jetty:                   9
jetty:                   9.3.8.v20160314  
jetty:                   9.3.8.v20160314-1 
jetty:                   9.3.8.v20160314-1-2016-09-16_04_33  
```

meltsufin:
@gregw For Jetty, don't we want a tag like `jetty: 9.3` as well?

gregw:
@joakime probably.   I've added to the branch and created the PR now: https://github.com/GoogleCloudPlatform/jetty-runtime/pull/44

meltsufin:
I'm thinking we can eliminate the `project.version` part altogether from the image version tags, if we're relying on time-stamps there.

joakime:
Don't get rid of the `${project.version}` specific docker tag, that's the most useful one for coordinating with other runtime builds, such as https://github.com/GoogleCloudPlatform/jetty-runtime

pure timestamp tags are fine for manual builds and testing of bugs though, so I'm happy to continue seeing that as an option.

meltsufin:
@joakime can't we just rely on `:latest` for those chained CI builds, assuming the builds are scoped to a specific project anyway?
Generally, I think, Jetty image will depend on a time-stamped version of the openjdk image so that we can get reproducible builds.

joakime:
If we have multiple branches active, then `:latest` has no meaning anymore.

We also have non-docker artifacts, from this openjdk-runtime build, in the maven repository system with the same `${project.version}` identifier.

It would be easiest to tie together the openjdk (branch) --> jetty-runtime (branch) via the build's version identifier for both the docker and non-docker artifacts.

gregw:
@meltsufin this has been updated as we discussed and this and the jetty-runtime are hopefully good to merge.     

As we get closer to release, I think we should consider the profiles, also perhaps dropping the jetty version timestamp version... not sure what value that adds other than pedantics.

gregw:
PRs currently build the following tags:
```
REPOSITORY                          TAG                                           IMAGE ID            CREATED             SIZE
jetty                               9                                             bca783d7c26c        3 minutes ago       618.1 MB
jetty                               9.3                                           bca783d7c26c        3 minutes ago       618.1 MB
jetty                               9.3.8                                         bca783d7c26c        3 minutes ago       618.1 MB
jetty                               9.3.8-2016-09-22_00_25                        bca783d7c26c        3 minutes ago       618.1 MB
jetty                               9.3.8.v20160314                               bca783d7c26c        3 minutes ago       618.1 MB
jetty                               9.3.8.v20160314-2016-09-22_00_25              bca783d7c26c        3 minutes ago       618.1 MB
jetty                               latest                                        bca783d7c26c        3 minutes ago       618.1 MB
jetty                               9.3.8.v20160314-2016-09-22_00_22              bac5c81fd0b0        7 minutes ago       618.1 MB
openjdk                             8                                             93ed8fd58c74        11 minutes ago      591.4 MB
openjdk                             8u102                                         93ed8fd58c74        11 minutes ago      591.4 MB
openjdk                             8u102-2016-09-22_00_14                        93ed8fd58c74        11 minutes ago      591.4 MB
openjdk                             latest                                        93ed8fd58c74        11 minutes ago      591.4 MB
```

gregw:
@joakime it is a good point that you make about non-docker artefacts.   
Is it possible to use the datestamp as the maven version?   

Perhaps we can have a version number like 2016.9.22.0.22 that matches a tag of 2016-09-22_00_22?

Anyway, I'd still like to get this PR merged now-ish so we at least get close to the final names/tags and we can open another issue to discuss how maven artefacts will be versioned and any profiles that we may need.

meltsufin:
@gregw I don't think we need the minor version parts like `v20160314` in jetty and `u102` in openjdk version tags because we will only release one minor version at a time. Also, I think we need to encode the JRE version somehow in our tags. I think the list of tags should be as follows:

```
REPOSITORY                          TAG                                           IMAGE ID            CREATED             SIZE
jetty                               9                                             bca783d7c26c        3 minutes ago       618.1 MB
jetty                               9.3                                           bca783d7c26c        3 minutes ago       618.1 MB
jetty                               9.3.8                                         bca783d7c26c        3 minutes ago       618.1 MB
jetty                               9-jre8                                        bca783d7c26c        3 minutes ago       618.1 MB
jetty                               9.3-jre8                                      bca783d7c26c        3 minutes ago       618.1 MB
jetty                               9.3.8-jre8                                    bca783d7c26c        3 minutes ago       618.1 MB
jetty                               9.3.8-jre8-2016-09-22_00_25                   bca783d7c26c        3 minutes ago       
jetty                               latest                                        bca783d7c26c        3 minutes ago       618.1 MB
openjdk                             8                                             93ed8fd58c74        11 minutes ago      591.4 MB
openjdk                             8-2016-09-22_00_14                            93ed8fd58c74        11 minutes ago      591.4 MB
openjdk                             latest                                        93ed8fd58c74        11 minutes ago      591.4 MB
```

joakime:
Note: the official openjdk image from hub.docker.com has the "u102" portion on the tags.

See: https://hub.docker.com/_/openjdk/

gregw:
@meltsufin the openjdk `u102` and the jetty `v20160314` are not equivalent.  The jetty date stamp is indeed entirely redundant information and only included in our release version for procedural stuff. I definitely agree that it can be dropped.

The `u102` on the other hand is very important version information and there are definite dependencies on specific versions of the JDK - eg the ALPN implementation is dependent on specific openjdk instances.  It is thus very important to be able to depend on that tag - to be more specific than any jre 8 version and less specific than a timestamped full version.

It is probably a good idea to include the openjdk version in the jetty image tag because it may be that users wish to upgrade to a new JVM without changing jetty version, or to bisect a bug between two JVM versions.   Thus I believe it should be:
```
REPOSITORY                          TAG
jetty                               9    
jetty                               9.3
jetty                               9.3.11
jetty                               9.3-8u102 
jetty                               9.3.11-8u102
jetty                               9.3.11-8u102-2016-09-22_00_25 
jetty                               latest
openjdk                             8
openjdk                             8u102
openjdk                             8u102-2016-09-22_00_14
openjdk                             latest
```

meltsufin:
Your point about update versions of JDK is well taken. The list looks good to me except for `jetty:9-8u102` tag missing.

meltsufin:
I think there is still a lot of uncertainty about what level of granularity the tags should be. In the most granular case, there are just too many possibilities to be manageable. In the interests of moving this PR forward, for the time being, we should just add these tags:
```
REPOSITORY                          TAG
jetty                               9    
jetty                               9-2016-09-22_00_25 
jetty                               latest
openjdk                             8
openjdk                             8-2016-09-22_00_14
openjdk                             latest
```
This does not provide a lot of flexibility in pinning to minor versions, but it does address the original issues #8 and #9.

joakime:
I'm -1 on the overly simplified tagging.  (simplified tagging is good goal, but not to this extreme)
I'm -0 on tags with timestamps (there's no value in it, you can just use the image hash if you really need to get that specific image)

Also, the /jetty-runtime/ needs the openjdk tags to include the update number `u102` in the tag, otherwise the ALPN + HTTP/2 layers will not work ([they are configured and tied to specific versions of Java](https://www.eclipse.org/jetty/documentation/current/alpn-chapter.html#alpn-versions)).  _(Note: this requirement goes away with Java 9+)_

joakime:
Reminder: Jetty versioning is ...

```
<servlet_support_level>.<major_version>.<minor_version>
```

So having a tag on just `jetty:9` isn't granular enough.

It needs to be *at least* `jetty:9.3` (Each of these are a major version bump in jetty: 9.2, 9.3, 9.4)

Configuration in 9.3 is not guaranteed to be backward compatible with 9.2 (for example)

Other examples:
 - 9.2 had SPDY
 - 9.3 dropped SPDY, bumped up the Java minimum requirement to Java 8, and introduced HTTP/2

meltsufin:
Regarding tags with timestamps, we may reserve the right to do security patches for timestamped versions. So, the image hash may be too specific.

meltsufin:
How about this for now.
```
jetty                               9.3    
jetty                               9.3-2016-09-22_00_25 
jetty                               latest
openjdk                             8
openjdk                             8-2016-09-22_00_14
openjdk                             latest
```

Pinning to `jetty:9` seems to add no value over `jetty:latest`

joakime:
Still need the `openjdk:8u102` and `openjdk:8u92` images references (those could be the base image name, making your proposed tags of those base images good), as those will be pinned on the /jetty-runtime/ side for the ALPN + HTTP/2 support.

meltsufin:
I don't think we'll ever build the openjdk:8u92 image. So, can jetty-runtime just depend on openjdk:latest for ALPN support?

joakime:
Lets go with the simplified openjdk image references you are proposing.

We'll open a new issue if the /jetty-runtime/ ALPN+HTTP/2 supports need more granularity.

meltsufin:
Closing in favor of continuing work in a project-local branch in #15.

PR #15, opened 2016-09-23
Improve docker image names and tags

### LINKED ISSUE #8: Default docker image version label should contain a unique build id

meltsufin:
Fixes #8 and #9.

Moving #10 to a local branch.

### DIFF
--- a/logging/pom.xml
+++ b/logging/pom.xml
@@ -21,7 +21,7 @@
   <parent>
     <groupId>com.google.cloud.runtimes</groupId>
     <artifactId>openjdk-parent</artifactId>
-    <version>0.1.0-SNAPSHOT</version>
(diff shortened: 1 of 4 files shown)


meltsufin:
@aslo PTAL

aslo:
LGTM - I think the simpler tag list is fine for the time being

gregw:
Note that thinks like ALPN depend on a specific version of the JRE and will not work with `latest`.  If the JRE is updated then we often have to update the ALPN release to match - so there is no automatic upgrade path.  This will be better in JRE9, where ALPN will be supported natively.

But even then you want the ability to depend on a specific JRE version so you can work on JRE specific debugging.

meltsufin:
Tagging scheme discussion should continue here: #16.

PR #26, opened 2016-11-17
Switch to headless JVM

### LINKED ISSUE #23: Consider using package openjdk-8-jre-headless instead
The openjdk-8-jre-headless seems like a good fit for a docker image (which obviously has no framebuffer or UI).

meltsufin:
Fixes #23.

Note that the old java-compat image also used headless JVM. Also, this saves around 50Mb of space on the image.

### DIFF
--- a/openjdk8/src/main/docker/Dockerfile
+++ b/openjdk8/src/main/docker/Dockerfile
@@ -25,7 +25,7 @@ RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/a
  && apt-get -q update \
  && apt-get -y -q --no-install-recommends install \
     ca-certificates \
-    openjdk-8-jre=${openjdk8.version}'-*' \
+    openjdk-8-jre-headless=${openjdk8.version}'-*' \
     netbase \
     wget \ 
     unzip \


aslo:
:+1: SGTM

gregw:
LGTM, but we should also look at using -server

ludoch:
IIRC, would that cause regression for folks using awt classes (for ex to
generate PDF docs?)

gregw:
@ludoch my understanding of -server is that is just uses a different version of the JIT and GC configuration.  The AWT classes are still available.   However, I will test and evaluate that separately.

.. or are you saying this headless change could cause the regression?   I'll check that as well.

joakime:
@ludoch the -headless variant still have the awt/swing classes.

IIRC, It does 2 things ...

1. sets `-Djava.awt.headless=true` by default
2. doesn't ship with implementation of awt, such as `awt-x11`

The image this project currently produces also has no x11 components currently installed (even before this change).

Example: if a PDF generation library attempts to use a font (via `Font.createFont(...)`) on the current image, then it will likely fail, because there's no x11, no fontlib, no fonts (hence no font cache), etc ...

If you decide to stick with the full JVM, then you can optionally use the `-Djava.awt.headless=true` system property to set various `java` components into the headless mode as well.

For more information: http://www.oracle.com/technetwork/articles/javase/headless-136834.html

joakime:
One extra note:  if the project has a need for loading a Font via awt, then it will likely need to customize the Image anyway (to add `fontlib` package at a minimum, and optionally the specific Fonts it needs/wants to load). 

Is this kind of requirement in the scope of the default/shipped `openjdk-runtime`?

PR #34, opened 2016-12-15
Introduce app.jar convention for child containers [EXPERIMENTAL]

meltsufin:
I'm just putting this up for comment. Experimenting with `ONBUILD` to potentially simplify the creation of child containers that use `app.jar` as the convention for application archive.
Currently, users write a custom Dockerfile like this:
```
FROM gcr.io/google-appengine/openjdk
ADD springboot-petclinic-1.4.2.jar /app/app.jar
CMD ["java", "-jar", "/app/app.jar"]
```

As long as they rename the archive to `app.jar`, it would allow them to write the Dockerfile with just the FROM line:
```
FROM gcr.io/google-appengine/openjdk
```

One downside I can think of is that the `ONBUILD ADD app.jar /app/` command will be inherited by our jetty-runtime, and although harmless, would just be an unnecessary layer in the image.


jmcc0nn3ll:
I am a fan of _Convention_ over _Configuration_ in general, seems fine to me.

gregw:
@meltsufin is it harmless?    if app.jar does not exist when building subsequent images then you get an error.

Also, the openjdk-runtime may be used by many other java containers (like jetty) that do not start with an executable jar.  In fact, springboot is probably in the minority of java apps in having an executable jar.

So not a fan.

meltsufin:
@gregw I didn't realize it would actually fail the docker build if the jar file is not present. So, you're right, this won't work.

I still kind of want to set the working directory to simplify the writing of the custom Dockerfiles.

meltsufin:
I removed the 'ADD app.jar'. So now, a custom Dockerfile would look like this:
```
FROM gcr.io/google-appengine/openjdk
ADD springboot-petclinic-1.4.2.jar app.jar
```
Note that app.jar is still convention because it's in the default CMD.

Thoughts?

openjdk8/src/main/docker/Dockerfile
```diff
@@ -42,5 +42,7 @@ RUN tar Cxfvz /opt/cdbg /opt/cdbg/cdbg_java_agent.tar.gz --no-same-owner \
  && mkdir /var/log/app_engine \
  && chmod go+rwx /var/log/app_engine
 
+WORKDIR /app
+
 ENTRYPOINT ["/docker-entrypoint.bash"]
-CMD ["java","-version"]
+CMD ["java", "-jar", "app.jar"]
```
gregw: you are missing an EOL here

gregw:
LGTM.  The previous default of running `-version` was pretty useless, so this default CMD is at least useful for some an replaceable for others.

meltsufin:
@gregw I'm concerned about the `WORKDIR /app` instruction because it will actually create the `/app` directory. This may have an unexpected impact on the jetty image, where we try to create a symlink for `/app`. WDYT?

gregw:
Ah good point.   A "java -jar /app/app.jar" may be better and derived images can set the working directory.  I think this further reduces the users that will be able to use this, as those that want a working directory set will probably have their executable jar in that directory, but then they are free to make the directory /app for all.

Note also I'm not a huge fan of just `/app` and `/opt` or `/var` may be better places to encourage installation of the application?

meltsufin:
I remove the working directory setting, but kept the default command the same.
If someone does this:
```
FROM gcr.io/google-appengine/openjdk
ADD springboot-petclinic-1.4.2.jar app.jar
```
It will work, but they'll be adding the app.jar to `/`
They can also choose to do this:
```
FROM gcr.io/google-appengine/openjdk
WORKDIR /app
ADD springboot-petclinic-1.4.2.jar app.jar
```
and it should work as well.

PR #40, opened 2017-01-14
Java agent for heap info, thread dump logging @ VM shutdown 

jabubake:


jabubake:
- There has been a minor change in the logging requirements. I'll update and resend pull request. 
Closing current one.

PR #44, opened 2017-01-20
Issue #43 export env

### LINKED ISSUE #43: Add `env JAVA_OPTS="$JAVA_OPTS"` to `exec "$@"`
At this line: https://github.com/GoogleCloudPlatform/openjdk-runtime/blob/master/openjdk8/src/main/docker/docker-entrypoint.bash#L21

Maybe we change it to `exec env JAVA_OPTS="$JAVA_OPTS" "$@" 2>&1`? (I'm not sure if `2>&1` is needed. Suggestion welcomed).

I was using Scala Playframework, and JAVA_OPTS wasn't set properly. I have to make my own docker-entrypoint.bash.

Making this change would allow me to get rid of my own docker-entrypoint.bash.

gregw:
Fixes #43
Always source setenv.sh
export all env variables set


openjdk8/src/main/docker/setup-env.bash
```diff
@@ -29,4 +29,14 @@ fi
 : ${JAVA_OPTS:=-showversion ${JAVA_TMP_OPTS} ${DBG_AGENT} ${JAVA_HEAP_OPTS} ${JAVA_GC_OPTS} ${JAVA_GC_LOG_OPTS} ${JAVA_USER_OPTS}}
 
 
-
+export \
```
meltsufin: Can we add `export` inline where the variables are defined?

meltsufin: @gregw ping

gregw:
@meltsufin I think this is a less bash-magic way of doing it and keeps the exports in line.  Only lightly tested so far.

PR #47, opened 2017-01-24
Fix XML formatting

meltsufin:
Fixing indentation in the poms.


meltsufin:
:+1: on the static analysis. Do you mind creating an issue for that?

PR #49, opened 2017-01-25
Issue https://github.com/GoogleCloudPlatform/jetty-runtime/issues/120

gregw:
Make docker-entrypoint easily extensible for https://github.com/GoogleCloudPlatform/jetty-runtime/issues/120


meltsufin:
@gregw Thinking about this a bit more, I think I'd rather just have the child image append to the `setup-env.bash`, and then this PR is not needed.

gregw:
OK that works.

PR #52, opened 2017-02-01
Update JRE 8 to u121

meltsufin:


openjdk8/src/main/docker/Dockerfile
```diff
@@ -24,7 +24,7 @@ ENV GAE_IMAGE_LABEL ${docker.tag.prefix}${openjdk.version}-${maven.build.timesta
 RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/apt/sources.list.d/jessie-backports.list \
  && apt-get -q update \
  && apt-get -y -q --no-install-recommends install \
-    ca-certificates \
+    ca-certificates-java=20161107'*' \
```
meltsufin: Without this change, I can't get the build to succeed. I'm not sure why we were using `ca-certificates` rather than `ca-certificates-java` though.

gregw: That was an inherited install, so I'm not sure either.  `ca-certificates-java` depends on `ca-certificates` so this change should not break anything.

PR #53, opened 2017-02-02
Master gae zone

gregw:
Set the GAE_ZONE environment variable if it is not already set and the GAE_INSTANCE variable is set.


gregw:
This has #52 merged in anticipation

gregw:
See discussion in https://github.com/GoogleCloudPlatform/jetty-runtime/pull/81

PR #54, opened 2017-02-09
Cloud container builds

aslo:
Allows the use of [cloud container builder](https://cloud.google.com/container-builder/docs/) to perform docker builds in the cloud.

BEFORE MERGING
- [x] determine whether the `gcr.io/cloud-builders/maven` build step is suitable for our use


aslo:
@meltsufin Do you want to merge this now? Or wait until we figure out what we're doing with `gcr.io/cloud-builders/maven`?

scripts/cloudbuild.sh
```diff
@@ -0,0 +1,31 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+projectRoot=`dirname $0`/..
+
+DOCKER_NAMESPACE='gcr.io/$PROJECT_ID'
```
meltsufin: Does Cloud Build automatically substitute PROJECT_ID?

scripts/cloudbuild.sh
```diff
@@ -0,0 +1,31 @@
...
+RUNTIME_NAME="openjdk"
+CANDIDATE_NAME="8-`date +%Y-%m-%d_%H_%M`"
+
+export IMAGE="${DOCKER_NAMESPACE}/${RUNTIME_NAME}:${CANDIDATE_NAME}"
+echo "IMAGE: $IMAGE"
+
+envsubst < $projectRoot/cloudbuild.yaml.in > $projectRoot/target/cloudbuild.yaml
```
meltsufin: Is there any reason why you wouldn't want to do all of this in maven instead with maven-exec-plugin? You could also then avoid using `evnsubst` and do maven source filtering instead.

scripts/cloudbuild.sh
aslo: Yep, that's why the right side of this assignment is in single quotes, so $PROJECT_ID won't be evaluated locally. 

https://cloud.google.com/container-builder/docs/api/build-requests#substitutions

scripts/cloudbuild.sh
aslo: That's definitely an option. One nice thing about this approach is that it doesn't require maven (or java, for that matter) to be installed on the machine that kicks off the build. That would make it easier to run this in CI environments.

aslo: One other thing - The way you're proposing would look like
```
local maven -> cloud builder -> remote maven -> remote docker
```
It seems a bit weird to me to have the same maven project invoked during several different points in the same build, no?

meltsufin: Good point. Plus, there isn't really anything to build with Maven if you're using Cloud Build.

PR #56, opened 2017-02-16
Fix an issue where cloudbuild.sh fails to run

DominicKramer:
In particular, on a fresh checkout, running `scripts/cloudbuild.sh` would fail because a `target` directory
doesn't exist in the project root directory.  This is probably because `git` doesn't track empty directories.


DominicKramer:
I doesn't look like I have the required permissions to merge the PR.

PR #58, opened 2017-02-17
Fix difference between GAE_IMAGE_LABEL and actual image tag when released

meltsufin:


meltsufin:
@aslo PTAL

scripts/cloudbuild.sh
```diff
@@ -20,7 +20,7 @@ projectRoot=`dirname $0`/..
 
 DOCKER_NAMESPACE='gcr.io/$PROJECT_ID'
 RUNTIME_NAME="openjdk"
-CANDIDATE_NAME="8-`date +%Y-%m-%d_%H_%M`"
+export CANDIDATE_NAME="8-`date +%Y-%m-%d-%H-%M`"
```
gregw: Why is it `CANDIDATE_NAME` rather than `LABEL`?    Other than that, this is good and builds fine for me.

aslo: :+1: maybe it makes sense to change this to `LONG_TAG` to match what's in the pom?

pom.xml
```diff
@@ -28,12 +28,14 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <maven.build.timestamp.format>yyyy-MM-dd_HH_mm</maven.build.timestamp.format>
+    <maven.build.timestamp.format>yyyy-MM-dd-HH-mm</maven.build.timestamp.format>
```
aslo: Just curious, what's the reason for changing the underscores to hyphens here?

meltsufin: It's for consistency with jetty-runtime mostly.

scripts/cloudbuild.sh
meltsufin: I'll go with `DOCKER_TAG_LONG` to match terminology in POM.

PR #61, opened 2017-02-23
Tagging improvements when using cloudbuild

aslo:
This PR makes a few improvements to the way we handle tags when using cloud build.

* removes the `-Ddocker.tag.long=${DOCKER_TAG_LONG}` maven argument when using cloud build. (This argument doesn't do anything since we explicitly disable the maven docker build profile).
* ~~Tags the built image with the short tag, as well as the long tag. This is to have parity between the maven-built docker tags (for local development) and the cloudbuild docker tags.~~

### DIFF
--- a/cloudbuild.yaml.in
+++ b/cloudbuild.yaml.in
@@ -4,13 +4,13 @@
 steps:
 # Build the maven project, omitting the docker build step
 - name: 'gcr.io/cloud-builders/maven'
-  args: ['mvn', '-P-local-docker-build', 'clean', 'install', '-Ddocker.tag.long=${DOCKER_TAG_LONG}']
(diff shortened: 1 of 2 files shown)


meltsufin:
My concern with the short tag is that it will potentially retag an existing image, similarly to how we retag to `:latest` using the special "release" job. I think we want to keep it as a separate step.

aslo:
Hmm, good point. Do you think it still makes sense to have the short tag in the maven docker build?

meltsufin:
It's kind of convenient sometimes for local testing and just illustrating what all of the final tags would be, but I don't feel strongly about it.

aslo:
Same. Let's just leave it for now. I'll revert that part of this PR

PR #65, opened 2017-02-28
Use maven properties for cloud build

aslo:
Reuses maven properties in the cloud builder pipeline, so we don't need to define our tag formats in more than one place. See also https://github.com/GoogleCloudPlatform/jetty-runtime/pull/135 for more discussion.

### DIFF
--- a/scripts/cloudbuild.sh
+++ b/scripts/cloudbuild.sh
@@ -17,11 +17,19 @@
 set -e
 
 projectRoot=`dirname $0`/..
+buildProperties=$projectRoot/target/build.properties
+
+# reads a property value from a .properties file
+function read_prop {
+  grep "${1}" $buildProperties | cut -d'=' -f2
+}
+
+# invoke local maven to generate build properties file
+mvn generate-resources
 
 DOCKER_NAMESPACE='gcr.io/$PROJECT_ID'
 RUNTIME_NAME="openjdk"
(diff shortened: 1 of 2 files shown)


scripts/cloudbuild.sh
```diff
@@ -17,11 +17,19 @@
...
 projectRoot=`dirname $0`/..
+buildProperties=$projectRoot/target/build.properties
+
+# reads a property value from a .properties file
+function read_prop {
+  grep "${1}" $buildProperties | cut -d'=' -f2
+}
+
+# invoke local maven to generate build properties file
+mvn generate-resources
```
meltsufin: Try to run the execution directly with `mvn properties:write-project-properties@my-exec`.

scripts/cloudbuild.sh
```diff
@@ -17,11 +17,19 @@
...
 
 DOCKER_NAMESPACE='gcr.io/$PROJECT_ID'
 RUNTIME_NAME="openjdk"
-export DOCKER_TAG_LONG="8-`date -u +%Y-%m-%d-%H-%M`"
-
+DOCKER_TAG_LONG=$(read_prop "docker.tag.long")
```
meltsufin: Pass  `-Ddocker.tag.long=$DOCKER_TAG_LONG` back into maven in the first step of the cloud build so that the timestamps match in the environment variables we set.

PR #68, opened 2017-03-09
Adding App Engine Flexible and GKE instructions to Readme

### LINKED ISSUE #59: Document running this image on GKE
We should document it like this: https://github.com/GoogleCloudPlatform/nodejs-docker#container-engine--other-docker-hosts

meltsufin:
Fixes #59

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ write the Dockerfile like this:
 
 ```dockerfile
 FROM gcr.io/google-appengine/openjdk
-ADD your-application.jar app.jar
+COPY your-application.jar app.jar
 ```
       
 That will add the JAR in the correct location for the Docker container.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,39 @@
...
+env: flex
+```
+The runtime image `gcr.io/google-appenine/openjdk` will be automatically selected if you are attempting to deploy a JAR (`*.jar` file).
+
+If you want to use the image as a base for a custom runtime, you can specify `runtime: custom` in your `app.yaml` and then
+write the Dockerfile like this:
+
+```dockerfile
+FROM gcr.io/google-appengine/openjdk
+ADD your-application.jar app.jar
```
aslo: I couldn't help but notice that you used `ADD` here and `COPY` below in the GKE example. According to docker's [best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy) `COPY` is preferred.

meltsufin: Good catch!

PR #73, opened 2017-03-24
platform specific env setup

gregw:
This is a bit of a thought bubble....  however I think it would greatly simplify extending the environment setup if we had:
 a) a directory based mechanism so that scripts could be added rather than appended; 
 b) a platform specific script that would select other directories of scripts based on a single platform selector

This PR achieves this, however the platform selection algorithm  for local, gae, aws and heroku is rather rudimentary and needs to be extended to support additional flavours of GCP.

However, with a mechanism like this, it would be much easier to create images that will run anywhere without each mechanism repeating the platform detection heuristics.

Currently only the cloud debugger is in a platform specific location, however the jetty-runtime image would also benefit by dropping in the configuration for its `gae` module into only the `platform-gae.d` directory.


meltsufin:
This is an interesting idea, but I don't think we're ready to put this kind of structure in place before we have a better idea of what will go into each of these environment files. For the time being, I would rather keep it simple, though less "organized".  See: #74. We should later do a refactoring like what you're suggesting here.

gregw:
@meltsufin you are right that putting in a complex generic structure based on an example of 1 is overkill for now.  However, I am still drawn to the simplicity and flexibility of a `setup-env.d` that is added to or deleted from rather than a `setup-env.bash` that is appended to or replaced.

So I've updated to a halfway solution between what was originally proposed and your #74. What do you think?

meltsufin:
@gregw The halfway solution might work. Did you forget to push the change though?

gregw:
@meltsufin oops sorry. Pushed now.

openjdk8/src/main/docker/setup-env.d/10-setup-env.bash
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+isTrue() {
```
meltsufin: Let's get rid of this method.

openjdk8/src/main/docker/setup-env.d/10-setup-env.bash
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
meltsufin: I would rename this file to `10-platform-env.bash`

openjdk8/src/main/docker/docker-entrypoint.bash
```diff
@@ -1,8 +1,12 @@
 #!/bin/bash
 
 # source the supported feature JVM arguments
```
meltsufin: Change comment to reflect the block bellow

openjdk8/src/test/resources/structure.yaml
```diff
@@ -35,9 +35,9 @@ fileExistenceTests:
   path: '/docker-entrypoint.bash'
   isDirectory: false
   shouldExist: true
-- name: 'Env setup script exists'
-  path: '/setup-env.bash'
-  isDirectory: false
+- name: 'Env setup scripts exists'
```
meltsufin: Let's add the new files in here for testing as well.

gregw:
@meltsufin Great! when will this be available as a build from gcr.io ?

meltsufin:
@aslo Do you want to try cutting a release using the new pipeline?

aslo:
Sure, I'll give it a go today.

meltsufin:
@gregw  It's available now: `gcr.io/google-appengine/openjdk:8-2017-04-12_14_02`

PR #75, opened 2017-03-28
Integration testing

aslo:
Implements support for the runtimes_common integration testing framework for openjdk-runtime. 

Note that only a subset of the test endpoints are implemented in this change:
- [serving](https://github.com/GoogleCloudPlatform/runtimes-common/tree/master/integration_tests#serving-root)
- [error reporting](https://github.com/GoogleCloudPlatform/runtimes-common/tree/master/integration_tests#error-reportingexceptions)

tests still to come:
- [logging](https://github.com/GoogleCloudPlatform/runtimes-common/tree/master/integration_tests#standard-logging)
- [monitoring](https://github.com/GoogleCloudPlatform/runtimes-common/tree/master/integration_tests#monitoring)


aslo:
@meltsufin FYI that above CLA message was caused by a bad rebase that I did earlier - the git history appears to be fixed now, and only my commits are in this branch, so I think that message no longer applies.

test-application/src/main/resources/logging.properties
```diff
@@ -0,0 +1,9 @@
+.level=INFO
```
meltsufin: I would remove it or put a comment in it for now that it's not used.

scripts/cloudbuild.sh
```diff
@@ -43,5 +44,8 @@ then
   curl -s https://raw.githubusercontent.com/GoogleCloudPlatform/python-runtime/master/scripts/local_cloudbuild.py | \
   python3 - --config=$projectRoot/target/cloudbuild_local.yaml --output_script=$projectRoot/target/cloudbuild_local.sh
 else
+  # invoke cloud container builder to build the image
   gcloud container builds submit --config=$projectRoot/target/cloudbuild.yaml .
+  # run integration tests on the built image
+  source $dir/integration_test.sh $IMAGE
```
aslo: I actually think it might just make more sense to not call the integration_test script directly from the cloudbuild script. I think that's better than implementing a --skip-integration-test arg to the cloudbuild script. 

I'll also modify integration_test.sh to accept a `--local` arg.

scripts/integration_test.sh
```diff
@@ -0,0 +1,54 @@
...
+pushd $testAppDir
+mvn clean install
+popd
+
+# deploy to app engine
+pushd $deployDir
+export STAGING_IMAGE=$imageUnderTest
+envsubst < Dockerfile.in > Dockerfile
+echo "Deploying to App Engine with dockerfile:"
+cat Dockerfile
```
aslo: good point, I'll just remove that line since it just adds more noise to the output

aslo:
@meltsufin I believe all of your requested changes are in. Mind taking another look?

DEVELOPING.md
```diff
@@ -16,11 +16,27 @@ The resulting image is called openjdk
...
+$ ./scripts/cloudbuild.sh gcr.io/$PROJECT_ID/openjdk --local
+```
+
+# Running Tests
+Integration tests can be run via [Google Cloud Container Builder](https://cloud.google.com/container-builder/docs/overview).
+These tests deploy a sample test application to App Engine using the provided runtime image, and 
+exercise various integrations with other GCP services.
+```bash
+$ PROJECT_ID=my-project
+$ ./scripts/integration_test.sh gcr.io/$PROJECT_ID/openjdk
```
meltsufin: This normally will not work, even after doing a cloud build first because the `:latest` tag will not be available. Maybe you should say, first find the image you want to test against by using
```
gcloud beta container images list-tags gcr.io/$PROJECT_ID/openjdk
```

DEVELOPING.md
```diff
@@ -16,11 +16,27 @@ The resulting image is called openjdk
...
+```
+
+To drive the integration tests using local cloud build, pass in the `--local` argument.
```
meltsufin: It seems like there's little utility to the local run because the image under test still has to be on GCR.
I wonder if there is any way around it to support the case of running integration tests on the local image.

aslo: I don't really think there's a good way around it. At the end of the day, the tests need to make requests to various stackdriver APIs, so it's not really "local" even if the other portions are. 

I'm cool with removing the local run option - each additional configuration adds more cognitive load, and if it's not delivering much value, let's remove it.

meltsufin: Agreed, let's drop it `--local` from integration testing. It sill makes sense for the image build though.

scripts/integration_test.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# build the test app
+pushd $testAppDir
+mvn clean install
+popd
+
+# deploy to app engine
+pushd $deployDir
+export STAGING_IMAGE=$imageUnderTest
+envsubst < Dockerfile.in > Dockerfile
```
meltsufin: Can you get rid of the `envsubst` here, and instead pass image under test as a property to `mvn clean install` above, and use resource filtering there?

aslo: I'm actually not crazy about that idea - is the motivation that you want to make this work across-platform? That wasn't really one of my goals with this PR, and if that was something we wanted to do, I'd probably want to rethink this bash-heavy approach to running these tests. 

Also, at some point in the near future, the integration test framework will be able to deploy the app being tested, and so we would need to go back and undo this change. (because the framework would need to do the templating of the `${STAGING_IMAGE}` variable).

meltsufin: I see, it's fine then.

PR #77, opened 2017-03-30
CI Changes

aslo:
Improvements to build scripts to enable releasing from a ci-like environment, where configuration is passed as environment variables. 

A few changes to note:
- addition of `ci-build.sh` which wraps the regular `build.sh` script (renamed from `cloudbuild.sh`)
- build scripts require both the `DOCKER_NAMESPACE` and the resulting `TAG` to be passed in, rather than using the image tag generated by maven


aslo:
@meltsufin Readme updated, PTAL

PR #86, opened 2017-04-19
Master entrypoint args

gregw:
The is a PR that brings down the `docker-entrypoint.bash` script from https://github.com/GoogleCloudPlatform/jetty-runtime/pull/155
and adds workspace unit tests for most of the entry point functionality


openjdk8/src/main/docker/docker-entrypoint.bash
```diff
@@ -1,23 +1,25 @@
 #!/bin/bash
 
+# If the first argument is the if full java command
```
meltsufin: Extra 'if'.

PR #90, opened 2017-05-01
Pin to integration test image

aslo:
We should pin to the specific image that we use for integration testing so that our tests don't get broken.


aslo:
I don't think they are publicly listed anywhere, but I found them by running `$ gcloud beta container images list-tags gcr.io/gcp-runtimes/integration_test`

PR #91, opened 2017-05-08
Use Kubernetes Downward API and cgroup to compute available memory on GKE

### LINKED ISSUE #14: Use Kubernetes Downward API to compute available memory on GKE
`GAE_MEMORY_MB` is not available on GKE, but the Kubernetes Downward API may be used to compute it. See: http://kubernetes.io/docs/user-guide/downward-api/

Related issue: #3

cassand:
Add mechanisms to determine the memory available for the container.

Previously the available memory was fetched from /proc/meminfo however those values are set for the supervisor and not the container which may lead to wrong heap allocation for the JVM.

This PR introduces two ways of computing the available memory:
* Through the Downward API (which requires the user to set environment variables in Kubernetes config files)
*  Using cgroup (This may not work for some containers e.g Rocket) 

The variable name for the Downward API is "KUBERNETES_MEMORY_LIMIT" which may be subject to change.

It might be useful to use cgroup only when some flags are set in order to conserve the current default behavior.

This PR fixes #14


meltsufin:
@cassand Can you please resolve the conflicts with master?

README.md
```diff
@@ -37,6 +37,33 @@ COPY your-application.jar app.jar
 You can then build the docker container using `docker build` or [Google Cloud Container Builder](https://cloud.google.com/container-builder/docs/).
 By default, the CMD is set to run the application JAR. You can change this by specifying your own `CMD` or `ENTRYPOINT`.
 
+In order to have an optimal usage of the container memory you can indicate a memory limit through the [Downward API](https://kubernetes.io/docs/tasks/configure-pod-container/environment-variable-expose-pod-information).
```
meltsufin: Add a subsection like `### Container Memory Limits`
"To help the runtime compute accurate JVM memory defaults when running on Kubernetes...."

README.md
```diff
@@ -37,6 +37,33 @@ COPY your-application.jar app.jar
...
+
+To do so add an environment variable named "KUBERNETES_MEMORY_LIMIT" with the value limits.memory and the name of your container.
```
meltsufin: Use code markup: `KUBERNETES_MEMORY_LIMIT` with the value `limits.memory`.

openjdk8/src/test/workspace/setup-java.bash
```diff
@@ -101,4 +101,19 @@ if [ "$TEST_MIN_HEAP" != "OK" -o "$TEST_MAX_HEAP" != "OK" ]; then
   exit 1
 fi
 
+
+#test GKE envrionment
+unset JAVA_OPTS TMPDIR GAE_MEMORY_MB HEAP_SIZE_MB JAVA_HEAP_OPTS JAVA_GC_OPTS JAVA_OPTS DBG_AGENT HEAP_SIZE_RATIO
+TMPDIR=/var/tmp
+KUBERNETES_MEMORY_LIMIT=20000000
```
meltsufin: Also specify `HEAP_SIZE_RATIO`, so that the test is not flaky, when/if we change the default ratio.

meltsufin:
@dlorenc Can you do quick pass at this PR. We decided to just put this in for the time being.

meltsufin:
@cassand @dlorenc  What do you think of renaming `KUBERNETES_MEMORY_LIMIT` to `CONTAINER_MEMORY_LIMIT_BYTES` and calculating it from `GAE_MEMORY_MB` by default, and then using `CONTAINER_MEMORY_LIMIT_BYTES` everywhere.

openjdk8/src/test/workspace/setup-java.bash
```diff
@@ -101,4 +101,20 @@ if [ "$TEST_MIN_HEAP" != "OK" -o "$TEST_MAX_HEAP" != "OK" ]; then
...
+HEAP_SIZE_RATIO=30
+source /setup-env.d/30-java-env.bash
+
+TEST_MIN_HEAP=$(echo $JAVA_OPTS | sed 's/.*-Xms5.*/OK/')
```
meltsufin: Can we be more precise here that `-Xms5.*`?

README.md
```diff
@@ -37,6 +37,34 @@ COPY your-application.jar app.jar
 You can then build the docker container using `docker build` or [Google Cloud Container Builder](https://cloud.google.com/container-builder/docs/).
 By default, the CMD is set to run the application JAR. You can change this by specifying your own `CMD` or `ENTRYPOINT`.
 
+### Container Memory Limits
+In order to have an optimal usage of the container memory you can indicate a memory limit through the [Downward API](https://kubernetes.io/docs/tasks/configure-pod-container/environment-variable-expose-pod-information).
```
meltsufin: Can you change the first sentence to say: "To help the runtime compute accurate JVM memory defaults when running on Kubernetes, you can indicate memory limit...."?

PR #92, opened 2017-05-08
Expose heap size as a percentage

### LINKED ISSUE #93: Set heap size as a percentage of the available memory
It might be more convenient for users to set the heap size as percentage of the container memory.

For example through an environment variable.

cassand:
Allow user to fix the percentage of the memory allocated to the heap.
This PR fixes #93


meltsufin:
@cassand Can you also create an issue for it and reference it in the PR description? Thanks!

aslo:
Nice work!

meltsufin:
@cassand very nice, and thanks for adding the test!!

Issue #64, opened 2017-02-28
cdbg_extra_class_path arguments are invalid for Jar based deployments


patflynn:
The image passes:
`--cdbg_extra_class_path=/webapps/root/WEB-INF/classes:/webapps/root/WEB-INF/lib`

as a fragment of:
```
java -showversion -agentpath:/opt/cdbg/cdbg_java_agent.so=--log_dir=/var/log/app_engine,--logtostderr=false,--cdbg_extra_class_path=/webapps/root/WEB-INF/classes:/webapps/root/WEB-
INF/lib -Xms491M -Xmx491M -XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+PrintCommandLineFlags -jar /app/spring-boot-simple-5.0.jar
```

which causes the /var/log/app_engine/cdbg_java_agent.INFO to contain:

```
cdbg_java_agent.8c8c13cfad57.invalid-user.log.INFO.20170228-155954.1  cdbg_java_agent.8c8c13cfad57.invalid-user.log.WARNING.20170228-155955.1  cdbg_java_agent.INFO  cdbg_java_agent.WARNING  custom_logs
root@8c8c13cfad57:/# cat /var/log/app_engine/cdbg_java_agent.INFO
Log file created at: 2017/02/28 15:59:54
Running on machine: 8c8c13cfad57
Log line format: [IWEF]mmdd hh:mm:ss.uuuuuu threadid file:line] msg
I0228 15:59:54.725625    15 jvmti_globals.cc:336] Build time: Jan  3 2017 09:10:03
I0228 15:59:54.725790    15 jvmti_agent.cc:157] Java debuglet initialization started
I0228 15:59:54.726048    15 jvmti_agent.cc:191] Java debuglet initialization completed
I0228 15:59:54.804965    15 jvmti_agent.cc:202] Java VM started
I0228 15:59:54.809967    15 jvmti_agent.cc:212] JvmtiAgent::JvmtiOnVMInit initialization time: 5007 microseconds
I0228 15:59:54.810044    26 jvmti_agent_thread.cc:99] Agent thread started: CloudDebugger_main_worker_thread
I0228 15:59:54.810212    26 jvm_internals.cc:376] Loading internals from /opt/cdbg/cdbg_java_agent_internals.jar
[...]

PR #94, opened 2017-05-09
Sets RUNTIME_DIR in runtimes, instead of openjdk base image.

joaoandremartins:
Helps fixing #64

Used in conjunction with https://github.com/GoogleCloudPlatform/jetty-runtime/pull/164

Will also require a change in format-env-appengine-vm.sh

### DIFF
--- a/openjdk8/src/main/docker/setup-env.d/20-debug-env.bash
+++ b/openjdk8/src/main/docker/setup-env.d/20-debug-env.bash
@@ -8,4 +8,5 @@ fi
 
 if [ "$DBG_ENABLE" = "true" ]; then
   unset CDBG_DISABLE
+  DBG_AGENT="$(/opt/cdbg/format-env-appengine-vm.sh)"
 fi
--- a/openjdk8/src/main/docker/setup-env.d/20-debug-env.bash
+++ b/openjdk8/src/main/docker/setup-env.d/20-debug-env.bash
@@ -8,5 +8,4 @@ fi
 
 if [ "$DBG_ENABLE" = "true" ]; then
   unset CDBG_DISABLE
-  DBG_AGENT="$( RUNTIME_DIR=$JETTY_BASE /opt/cdbg/format-env-appengine-vm.sh )"
 fi


openjdk8/src/main/docker/setup-env.d/20-debug-env.bash
```diff
@@ -8,5 +8,4 @@ fi
 
 if [ "$DBG_ENABLE" = "true" ]; then
   unset CDBG_DISABLE
-  DBG_AGENT="$( RUNTIME_DIR=$JETTY_BASE /opt/cdbg/format-env-appengine-vm.sh )"
```
meltsufin: This is completely removing the debugger. I don't think we want to do that.

openjdk8/src/main/docker/setup-env.d/20-debug-env.bash
```diff
@@ -8,5 +8,5 @@ fi
...
+  DBG_AGENT="$(/opt/cdbg/format-env-appengine-vm.sh)"
```
meltsufin: I guess it can now just be
`DBG_AGENT="/opt/cdbg/format-env-appengine-vm.sh"`

joaoandremartins: I think you still want `/opt/cdbg/format-env-appengine-vm.sh` to be evaled, no?

meltsufin: Right, never mind.

joaoandremartins: $ export DBG_AGENT="$(/usr/local/google/home/joaomartins/Downloads/cdbg_java_agent/format-env-appengine-vm.sh)"
$ echo $DBG_AGENT
-agentpath:/usr/local/google/home/joaomartins/Downloads/cdbg_java_agent/cdbg_java_agent.so=--log_dir=/var/log/app_engine,--logtostderr=false,--cdbg_extra_class_path=/webapps/root/WEB-INF/classes:/webapps/root/WEB-INF/lib

(without the format-env-appengine-vm.sh fix)

joaoandremartins:
This change is pending internal change 155509870 from being submitted and published.

PR #97, opened 2017-05-12
Add integration test for GKE

### LINKED ISSUE #96: Test the integration with Google Container Engine
We should add an integration test for the deployment to Google Container Engine (based on [Integration Frameworks](https://github.com/GoogleCloudPlatform/runtimes-common/tree/master/integration_tests))

cassand:
Add a script to setup environment for integration test on GKE using [runtimes-common integration testing framework](https://github.com/GoogleCloudPlatform/runtimes-common/tree/master/integration_tests)

The script does the following steps: 
- Build a test application into a docker image
- Upload the image to Google Container Registry
- Create a Kubernetes cluster on GKE
- Deploy the application
- Run the integration test

It assumes that kubectl executable is available.

This PR fixes #96

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -32,7 +32,7 @@ $ ./scripts/build.sh gcr.io/$PROJECT_ID $TAG --local
 
 # Running Tests
 Integration tests can be run via [Google Cloud Container Builder](https://cloud.google.com/container-builder/docs/overview).
-These tests deploy a sample test application to App Engine using the provided runtime image, and 
+These tests deploy a sample test application to App Engine and to Google Container Engine using the provided runtime image, and
 exercise various integrations with other GCP services. Note that the image under test must be pushed 
 to a gcr.io repository before the integration tests can run.
 ```bash
@@ -41,3 +41,14 @@ $ gcloud docker -- push $RUNTIME_IMAGE
 $ ./scripts/integration_test.sh $RUNTIME_IMAGE
 ```
 
+You also have the possibility to run the tests only on App Engine or only on Google Container Engine.
+
+* For App Engine:
+```bash
+$ ./scripts/ae_integration_test.sh $RUNTIME_IMAGE
+```
+
+* For Google Container Engine:
+```bash
+$ ./scripts/gke_integration_test.sh $RUNTIME_IMAGE
+```
--- a/scripts/gke_integration_test.sh
+++ b/scripts/gke_integration_test.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+# Copyright 2017 Google Inc. All rights reserved.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Setup environment for CI with GKE.
+# This include:
+# - Building a test application into a docker image
+# - Upload the image to Google Container Registry
+# - Create a Kubernetes cluster on GKE
+# - Deploy the application
+# - Run the integration test
+set -e
+
+readonly dir=$(dirname $0)
+readonly projectRoot="$dir/.."
+readonly testAppDir="$projectRoot/test-application"
+readonly deployDir="$testAppDir/target/deploy"
+
+readonly projectName=$(gcloud info \
+                | awk '/^Project: / { print $2 }' \
+                | sed 's/\[//'  \
+                | sed 's/\]//')
+readonly imageName="openjdk-gke-integration"
+readonly imageUrl="gcr.io/$projectName/$imageName"
+readonly clusterName="openjdk-gke-integration"
+
+readonly imageUnderTest=$1
+if [[ -z "$imageUnderTest" ]]; then
+  echo "Usage: ${0} <image_under_test>"
+  exit 1
+fi
+
+# build the test app
+pushd ${testAppDir}
+mvn clean install
+popd
+
+# deploy to app engine
+pushd ${deployDir}
+export STAGING_IMAGE=${imageUnderTest}
+envsubst < "Dockerfile.in" > "Dockerfile"
+export TESTED_IMAGE=${imageUrl}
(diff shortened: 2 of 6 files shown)


scripts/gke_integration_test.sh
```diff
@@ -0,0 +1,86 @@
...
+popd
+
+# deploy to app engine
+pushd ${deployDir}
+export STAGING_IMAGE=${imageUnderTest}
+envsubst < "Dockerfile.in" > "Dockerfile"
+export TESTED_IMAGE=${imageUrl}
+envsubst < "openjdk-spring-boot.yaml.in" > "openjdk-spring-boot.yaml"
+
+echo "Deploying image to Google Container Engine..."
```
meltsufin: "Deploying image to Google Container **Registry**..."

scripts/gke_integration_test.sh
```diff
@@ -0,0 +1,86 @@
...
+docker build -t "$imageName" .
+docker tag "$imageName" "$imageUrl"
+gcloud docker -- push gcr.io/${projectName}/openjdk-gke-integration
```
meltsufin: @aslo Should we do all these steps or use the builder instead? (In any case, this could be a separate PR.)

aslo: It's worth thinking about. One reason why it might not be a good idea - these tests require using a custom base image, which would require overriding the default builder to point to the base image under test. 

I think if we want to test openjdk runtime candidates with the builder pipeline, that would belong in the builder's integration tests.

scripts/integration_test.sh
```diff
@@ -14,40 +14,17 @@
...
-DEPLOYED_APP_URL="http://$(gcloud app describe | grep defaultHostname | awk '{print $2}')"
-echo "Running integration tests on application that is deployed at $DEPLOYED_APP_URL"
-
-# run in cloud container builder
-gcloud container builds submit \
-  --config $dir/integration_test.yaml \
-  --substitutions "_DEPLOYED_APP_URL=$DEPLOYED_APP_URL" \
-  $dir
+${dir}/ae_integration_test.sh ${imageUnderTest}
+${dir}/gke_integration_test.sh ${imageUnderTest}
```
meltsufin: Maybe add a final line here saying:
App Engine integration tests: PASSED/FAILED
Container Engine integration tests: PASSED/FAILED

DEVELOPING.md
```diff
@@ -38,6 +38,16 @@ to a gcr.io repository before the integration tests can run.
 ```bash
 $ RUNTIME_IMAGE=gcr.io/my-project-id/openjdk:my-tag
 $ gcloud docker -- push $RUNTIME_IMAGE
-$ ./scripts/integration_test.sh $RUNTIME_IMAGE
```
meltsufin: Let's leave this in for the case when people want to run both integration tests.
Also, in the the paragraph above, mention GKE in addition to App Engine.

DEVELOPING.md
```diff
@@ -38,6 +38,16 @@ to a gcr.io repository before the integration tests can run.
...
 ```
 
+You have the possibility to either run the tests on App Engine or on Google Container Engine.
```
meltsufin: Then here just say that you have the option to run the AE and GKE tests individually like this:

scripts/integration_test.sh
```diff
@@ -17,14 +17,29 @@
 # Setup environment for CI.
 set -e
 
-dir=`dirname $0`
+readonly dir=`dirname $0`
+
+GetStatusOfContainerBuild () {
```
meltsufin: Can we just rely on the exit code of the script instead? I believe it's 0 when pass, and non-zero when failed.

scripts/integration_test.sh
```diff
@@ -17,14 +17,29 @@
...
-${dir}/ae_integration_test.sh ${imageUnderTest}
-${dir}/gke_integration_test.sh ${imageUnderTest}
+AE_OUTPUT=$(${dir}/ae_integration_test.sh ${imageUnderTest})
+GKE_OUTPUT=$(${dir}/gke_integration_test.sh ${imageUnderTest})
+
+AE_STATUS=$(GetStatusOfContainerBuild "$AE_OUTPUT")
+GKE_STATUS=$(GetStatusOfContainerBuild "$GKE_OUTPUT")
+
+echo "App Engine integration tests: $AE_STATUS"
+echo "Container Engine integration tests: $GKE_STATUS"
```
meltsufin: Also, I think we want to make sure the script return non-zero exit code, if either of the tests failed.

scripts/integration_test.sh
```diff
@@ -14,40 +14,33 @@
...
+${dir}/ae_integration_test.sh ${imageUnderTest}
+AE_OUTPUT=$?
+AE_STATUS=$(GetStatusOfContainerBuild "$AE_OUTPUT")
+
+${dir}/gke_integration_test.sh ${imageUnderTest}
+GKE_OUTPUT=$?
+GKE_STATUS=$(GetStatusOfContainerBuild "$GKE_OUTPUT")
+
+echo "App Engine integration tests: $AE_STATUS"
+echo "Container Engine integration tests: $GKE_STATUS"
```
meltsufin: I think we need an `exit` instruction here that returns `0` if both passed or `1` otherwise.

aslo: I think that's not necessary, since the script has `set -e` at the top. It will exit immediately if any command exits with non-zero status. 

If that's not what @cassand intends here, then `set -e` should be removed and we can compute the exit code of this script based on how `ae_integration_test` and `gke_integration_test` finish.

meltsufin: I totally missed the `set -e` part. Thanks for pointing it out! I think we just want to disable `set -e` right before calling out to the test scripts by using `set +e`.

cassand: It seems fine that way, however, I am not sure that the command `gcloud container builds` exit with 1 if the remote build fail, in which case it may be useful to exit based on the status.

cassand: I confirm that the command `gcloud container builds` exit with a non zero exit code in case of failure

scripts/integration_test.sh
```diff
@@ -14,40 +14,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Setup environment for CI.
```
meltsufin: "Exit as soon as any of the commands in the script fail"

scripts/integration_test.sh
```diff
@@ -15,6 +15,8 @@
 # limitations under the License.
 
 # Setup environment for CI.
+# This script exit as soon as any of the commands in the integration scripts fail
```
meltsufin: exit -> exits

PR #102, opened 2017-05-16
Install kubectl during CI build

cassand:
Fix #102


scripts/gcloud-init.sh
```diff
@@ -45,3 +45,4 @@ export PATH=$DIR/google-cloud-sdk/bin:$PATH
 gcloud auth activate-service-account --key-file=$KEYFILE
 gcloud config set project $GCP_PROJECT
 gcloud components install beta -q
+gcloud components install kubectl
```
aslo: It might be cleaner to install all components on one line: `gcloud components install beta kubectl -q`

PR #110, opened 2017-06-06
Fix DBG_ENABLE checking and add tests

### LINKED ISSUE #107: Adding `DBG_ENABLE: true` to app.yaml disables Cloud Debugger
```yaml
env_variables:
   DBG_ENABLE: true
```

meltsufin:
Fixes #107

### DIFF
--- a/openjdk8/src/main/docker/setup-env.d/20-debug-env.bash
+++ b/openjdk8/src/main/docker/setup-env.d/20-debug-env.bash
@@ -1,12 +1,15 @@
 #!/bin/bash
 
 # configure Cloud Debugger
+
+export DBG_SCRIPT_PATH=${DBG_SCRIPT_PATH:-"/opt/cdbg/format-env-appengine-vm.sh"}
+
 if [ "$PLATFORM" = "gae" ]; then
   export DBG_AGENT=
-  export DBG_ENABLE=${DBG_ENABLE:-$( if [[ -z "${CDBG_DISABLE}" && -x /opt/cdbg/format-env-appengine-vm.sh ]] ; then echo true; else echo false ; fi )}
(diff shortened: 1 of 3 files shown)


openjdk8/src/main/docker/setup-env.d/20-debug-env.bash
```diff
@@ -1,12 +1,15 @@
...
+export DBG_SCRIPT_PATH=${DBG_SCRIPT_PATH:-"/opt/cdbg/format-env-appengine-vm.sh"}
+
 if [ "$PLATFORM" = "gae" ]; then
   export DBG_AGENT=
-  export DBG_ENABLE=${DBG_ENABLE:-$( if [[ -z "${CDBG_DISABLE}" && -x /opt/cdbg/format-env-appengine-vm.sh ]] ; then echo true; else echo false ; fi )}
+  export DBG_ENABLE=${DBG_ENABLE:-$( if [[ -z "${CDBG_DISABLE}" && -x "${DBG_SCRIPT_PATH}" ]] ; then echo true; else echo false ; fi )}
 fi
 
-if [ "$DBG_ENABLE" = "true" ]; then
+if [ "$DBG_ENABLE" = "true" -o "$DBG_ENABLE" = "True" ]; then
```
aslo: It might be a good idea to move this into a bash function called `isTrue` (and perhaps also create one called `isFalse`). I'd imagine that there will be more flags that are parsed from yaml that our bash scripts will have to act on.

meltsufin: Good idea! Currently we only use it in this particular case and I put the function in this file, but we can move it out to a common place if other cases come up.

PR #115, opened 2017-06-15
Thread Dump and Heap Info Shutdown Logs Support

### LINKED ISSUE #114: Shutdown Logging Support
**TL;DR: On a VM shutdown, allow logging the application’s thread dump and heap info to Cloud Logging.**

On App Engine, sometimes autohealer decides to kill unhealthy VMs. In that case we don't have sufficient logging done from the VM to debug if it was an issue with the app or the VM. This can happen when the app code gets stuck (like deadlock) and stopped returning requests, including health checks. 

Because it is possible that the JVM is ailing, it might not have enough resources to log to Cloud Logging directly. So we will be using stdout + fluentd to capture app container logs.

Explicit user intent is required for potentially sensitive information. The following environment variables will be used to enable/disable app container shutdown reporting (must be set to true or false):
 
`SHUTDOWN_LOGGING_THREAD_DUMP` - save thread traces
`SHUTDOWN_LOGGING_HEAP_INFO` - save heap information
 
The runtime will capture `SIGTERM`, and run debugging tools on the JVM to emit the thread traces and heap information to stdout.
 
Further, for customers with a small number of VMs we default to reporting shutdown logs for all VMs. For larger customers running hundreds or thousands of instances, we would allow them to be able to sample logs. This will be done by using the environment variable `SHUTDOWN_LOGGING_SAMPLE_THRESHOLD` which is an int between 0 and 100. 0 means no VMs report logs, 100 means all VMs report logs. (If this env var is not set, we default to reporting for all VMs). The runtime will then generate a random number between 0 (inclusive) and 100 (exclusive) and if it’s less than the threshold set, it should do the trap-and-log logic described above.

@karan

meltsufin:
Fixes #114


openjdk8/src/main/docker/setup-env.d/05-utils-env.bash
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
aslo: Small nit: I would probably just name this file `05-utils.bash`. Nothing in this file directly has to do with the environment.

openjdk8/src/main/docker/shutdown-wrapper.bash
```diff
@@ -0,0 +1,37 @@
...
+    jcmd $PID GC.class_histogram >> $HEAP_INFO_FILE 2>&1
+  fi
+}
+
+# capture the TERM signal and first generate the thread dump and/or heap info
+trap 'java_shutdown_hook; kill -TERM $PID' TERM
+$@ &
+PID=$!
+wait $PID
+wait $PID
```
aslo: Why do you need to do this twice?

openjdk8/src/test/workspace/entrypoint-shutdown-logging.bash
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+
+sed 's/exec /# /' /docker-entrypoint.bash > /tmp/entrypoint.bash
```
aslo: It would be great to not have to do this... If you moved this logic outside of the `docker-entrypoint.bash`, would you be able to test the source unmodified?

openjdk8/src/main/docker/shutdown-wrapper.bash
meltsufin: There's some explanation [here](http://veithen.github.io/2014/11/16/sigterm-propagation.html).
As far as I understand, the first wait is for when the JVM receives the signal, and the second is for when it terminates.

meltsufin:
@aslo @cassand Can you guys take another look? I added support for JVM running under a different user. I also add a local integration test.

openjdk8/src/main/docker/shutdown/shutdown-wrapper.bash
```diff
@@ -0,0 +1,44 @@
...
+
+function java_shutdown_hook {
+  JAVA_PID=`ps -C java -o pid --no-headers`
+  PUSER=`ps -o ruser --no-headers -p $JAVA_PID`
+
+  # send thread dump to a file
+  if is_true "${SHUTDOWN_LOGGING_THREAD_DUMP}"; then
+    export THREAD_DUMP_FILE=${THREAD_DUMP_FILE:-/app.shutdown.threads}
+    # run jcmd under JVM process owner and skip first 3 lines of output
+    su $PUSER -c "jcmd $JAVA_PID Thread.print" | tail -n +3 &> $THREAD_DUMP_FILE
```
cassand: Would the current user always have enough privilege to login as the owner of java ?

For example if the user is changed to `jetty` by the end user (Dockerfile) and the java executable is own by `root`.

scripts/local_integration_test.sh
```diff
@@ -0,0 +1,80 @@
...
+# build app container locally
+pushd $deployDir
+export STAGING_IMAGE=$imageUnderTest
+envsubst < Dockerfile.in > Dockerfile
+echo "Building app container..."
+docker build -t $APP_IMAGE .
+
+# run app container locally to test shutdown logging
+echo "Starting app container..."
+docker run --rm --name $CONTAINER -e "SHUTDOWN_LOGGING_THREAD_DUMP=true" -e "SHUTDOWN_LOGGING_HEAP_INFO=true" $APP_IMAGE &> $OUTPUT_FILE &
```
aslo: I'm a little uneasy about adding an additional test framework, but this is a good test and there isn't currently any other way to test this so I'm ok with it.

Also, make sure that you also add the execution as a buildstep in cloudbuild.yaml, since this will currently not be run by CI.

meltsufin: Good call about adding to `cloudbuild.yaml`!

meltsufin: @aslo wouldn't actually make more sense to add it to `ci-build.sh`?

meltsufin: ...Or rather `integration_test.sh`

aslo: Ah yep, that's right - I was thinking in the mindset of the jetty-runtime repo where everything is run in a single cloudbuild execution. Yeah, maybe integration_test.sh would be a good place to call this script.

PR #119, opened 2017-06-23
Set APP_DESTINATION env variable

aslo:
Sets a variable to show where web apps should be added in child containers.


openjdk8/src/main/docker/Dockerfile
```diff
@@ -47,4 +47,4 @@ RUN tar Cxfvz /opt/cdbg /opt/cdbg/cdbg_java_agent.tar.gz --no-same-owner \
 ENV APP_DESTINATION app.jar
 
 ENTRYPOINT ["/docker-entrypoint.bash"]
-CMD ["java", "-jar", $APP_DESTINATION]
+CMD ["java", "-jar", "app.jar"]
```
meltsufin: why?

aslo: CMD doesn't understand environment variables, so this wouldn't work the way it was. One option would be to use maven property variables to template in this value, which might be better.

PR #121, opened 2017-07-28
small improvements for development guide

balopat:
As I went through the build the first time (both cloud and local) I ran into two issues:
* my user did not have docker group setup - and that is assumed by the build script, so I mentioned that assumption explicitly in the guide 
* for the "local emulation" of the cloud build, a python script is pulled in from the python runtime - which is not an ideal solution necessarily - especially that the old version (the one I switched back to) worked, the new version doesn't, as it got modularized in [this commit ](https://github.com/GoogleCloudPlatform/python-runtime/commit/6f2275b44ff4a4b083f013f014fc37849952c79e) - I'll put in an issue to have a better fix than just depending on an older version

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -5,7 +5,7 @@ This document contains instructions on how to build and test this image.
 # Building the image
 
 ### Local build
-To build the image you need git, docker and maven installed:
+To build the image you need git, docker (your user needs to be part of the ``docker`` group to run docker without sudo) and maven installed:
 ```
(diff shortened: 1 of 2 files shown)


aslo:
👍 Thanks for the improvements to the build process! 

I'm also not crazy about the way we currently do local emulation of cloud builds. We can tackle this in a separate issue, but I might even prefer to remove it altogether, since maven is the recommended way to build locally. We've already done that in the Jetty repo. @meltsufin what do you think?

PR #123, opened 2017-07-29
adds structure test to ensure no java 7 runtime gets pulled in

### LINKED ISSUE #55: Verify that JRE 7 is not installed via ca-certificates-java
We're installing [ca-certificates-java](https://packages.debian.org/jessie-backports/ca-certificates-java) as a requirement for the openjdk, but it seems to depend on JRE 7. We need to make sure that we're not polluting the image with JRE 7.

balopat:
closes #55.

### DIFF
--- a/openjdk8/src/test/resources/structure.yaml
+++ b/openjdk8/src/test/resources/structure.yaml
@@ -11,8 +11,7 @@ commandTests:
 - name: 'ensure java7 is not installed as a dependency accidentally'
   setup: [[ 'chmod', '+x', '/workspace/openjdk8/src/test/workspace/test-no-java-7.bash' ]]
   command: [ '/workspace/openjdk8/src/test/workspace/test-no-java-7.bash' ]
-  expectedError: ['WARNING.*apt does not have a stable CLI interface yet.']
-  expectedOutput: ['OK.* is not installed']
+  excludedOutput: ['NOT OK.* is installed']
 - name: 'correct javac version is installed'
   command: ['javac', '-version']
   expectedError: ['javac 1\.${openjdk.version.major}.*${openjdk.version.minor}']
--- a/openjdk8/src/test/resources/structure.yaml
+++ b/openjdk8/src/test/resources/structure.yaml
@@ -8,9 +8,9 @@ commandTests:
 - name: 'correct java version is installed'
   command: ['java', '-version']
   expectedError: ['openjdk version "1\.${openjdk.version.major}.*${openjdk.version.minor}"']
-- name: 'ensure java7 is not installed as a dependency accidentally'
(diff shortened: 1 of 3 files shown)


openjdk8/src/test/resources/structure.yaml
```diff
@@ -8,6 +8,11 @@ commandTests:
 - name: 'correct java version is installed'
   command: ['java', '-version']
   expectedError: ['openjdk version "1\.${openjdk.version.major}.*${openjdk.version.minor}"']
+- name: 'ensure java7 is not installed as a dependency accidentally'
+  setup: [[ 'chmod', '+x', '/workspace/openjdk8/src/test/workspace/test-no-java-7.bash' ]]
+  command: [ '/workspace/openjdk8/src/test/workspace/test-no-java-7.bash' ]
+  expectedError: ['WARNING.*apt does not have a stable CLI interface yet.']
+  expectedOutput: ['OK.* is not installed']
```
aslo: I'm not sure that this will catch if java 7 is installed. I believe that the test will pass if the `expectedOutput` or `expectedError` are matched at least one time.

balopat: Agreed, thank you, for catching this! 

I tried it out with java 8 and it would have not caught it. 
I committed a new version that explicitly states "NOT OK. xxxx is installed" and now the test matches that as the ``excludedOutput``. I also removed ``expectedError`` as it is not really expected, just meant to be an explanatory piece - however it might just be more distracting than anything.

aslo:
Nice! Testing logic looks good. 

Did you consider running `apt-get update` before running the test? While building the openjdk docker container I believe we clear the apt caches, so it's possible that the `apt-cache dump`  command that you use to find java 7 packages might not be fully populated.

balopat:
I rewrote the check for an extensible blacklist checking test rather. 
The list of java/jre/jdk 7 packages are part of it now (maybe too much?). 
But this is much more understandable and safe. 
Thus, apt-get update is unnecessary. I also switched to ``dpkg``, which removed the annoying warning from the apt CLI.

openjdk8/src/test/workspace/test-blacklisted-pkgs.bash
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+BLACKLIST="jvm-7-avian-jre"
+BLACKLIST="java7-runtime-headless $BLACKLIST"
```
meltsufin: I think it's a bit verbose in this form.
Maybe just:
```sh
BLACKLIST="\
jvm-7-avian-jre \
libjsr107cache-java \
...
"
```

balopat: Done! Thank you!

PR #124, opened 2017-08-02
Adds monitoring integration tests

### LINKED ISSUE #76: Test stackdriver monitoring
Implement the /monitoring endpoint of the test application and implement the monitoring part of the test spec. See https://github.com/GoogleCloudPlatform/runtimes-common/tree/master/integration_tests#monitoring

balopat:
closes #76

### DIFF
--- a/scripts/gke_integration_test.sh
+++ b/scripts/gke_integration_test.sh
@@ -28,11 +28,20 @@ readonly projectRoot="$dir/.."
 readonly testAppDir="$projectRoot/test-application"
 readonly deployDir="$testAppDir/target/deploy"
 
+
+# The $TAG was introduced to be able to reuse the existing cluster but force redeployment.
+# Kubernetes' "kubectl apply -f" doesn't trigger a new deployment rollout unless there is a change in the yaml spec.
+# See https://github.com/kubernetes/kubernetes/issues/33664 for the debate around this behavior.
+
+if [ -z "${TAG}" ]; then
+  export TAG="$(date -u +%Y-%m-%d_%H_%M)"
+fi
+
 readonly projectName=$(gcloud info \
                 | awk '/^Project: / { print $2 }' \
                 | sed 's/\[//'  \
                 | sed 's/\]//')
-readonly imageName="openjdk-gke-integration"
+readonly imageName="openjdk-gke-integration:$TAG"
 readonly imageUrl="gcr.io/$projectName/$imageName"
 readonly clusterName="openjdk-gke-integration"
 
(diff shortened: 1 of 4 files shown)


test-application/src/main/java/com/google/cloud/runtimes/MonitoringTestController.java
```diff
@@ -0,0 +1,75 @@
...
+        return CreateTimeSeriesRequest.newBuilder()
+                .setNameWithProjectName(projectName)
+                .addTimeSeries(ts)
+                .build();
+    }
+
+    private String checkAndGetProjectIdFromEnv() {
+        String projectId = System.getenv("GOOGLE_CLOUD_PROJECT");
+        if (projectId == null) {
+            throw new IllegalArgumentException("Please set your project id! Usage: -DGOOGLE_CLOUD_PROJECT=YOUR_PROJECT_ID");
```
aslo: Does this mean that maven automatically converts arbitrary properties to environment variables?

balopat: well, yes, however I think this would be better to keep generic, e.g.: Please set your project id using the environment variable: GOOGLE_CLOUD_PROJECT
this change is coming up soon.

test-application/src/main/java/com/google/cloud/runtimes/MonitoringTestController.java
```diff
@@ -0,0 +1,75 @@
...
+                .addPoints(point)
+                .build();
+        return CreateTimeSeriesRequest.newBuilder()
+                .setNameWithProjectName(projectName)
+                .addTimeSeries(ts)
+                .build();
+    }
+
+    private String checkAndGetProjectIdFromEnv() {
+        String projectId = System.getenv("GOOGLE_CLOUD_PROJECT");
```
meltsufin: I would just use `ServiceOptions.getDefaultProjectId()` which will give you the same thing, but is more future/environment -proof.

scripts/gke_integration_test.sh
```diff
@@ -28,11 +28,15 @@ readonly projectRoot="$dir/.."
...
+if [ -z "${TAG}" ]; then
+  export TAG="$(date -u +%Y-%m-%d_%H_%M)"
+fi
+
 readonly projectName=$(gcloud info \
                 | awk '/^Project: / { print $2 }' \
                 | sed 's/\[//'  \
                 | sed 's/\]//')
-readonly imageName="openjdk-gke-integration"
+readonly imageName="openjdk-gke-integration:$TAG"
```
balopat: The $TAG was introduced to be able to reuse the existing cluster but force redeployment. Kubernetes' ``kubectl apply -f`` doesn't trigger a new deployment rollout unless there is a change in the yaml spec. See https://github.com/kubernetes/kubernetes/issues/33664 for the debate around this behavior.

meltsufin: Can you drop this comment into the code itself? Thanks!

balopat: sure, I added it.

PR #128, opened 2017-08-08
Common test application

aslo:
This PR makes a few changes to the test application:

- moves it from `/test-application` to `/java-runtimes-common/test-spring-application`. This implies that the test app is not specific to the openjdk runtime, but can be used to test others as well.
- implements new maven profiles and properties to customize the build:
  - `-Pruntime.custom` - profile for staging for deployment to a custom runtime (used for runtimes-common integration tests). The custom image must be provided using the `-Dapp.deploy.image` property. This profile is active by default.
  - `-Pruntime.java` - profile for staging for deployment to the default Java runtime
  - `-Dpackaging.type=jar` - packages the test app as a fat jar (default) 
  - `-Dpackaging.type=war` - packages the test app as a war (for use in the jetty/tomcat runtimes)


java-runtimes-common/test-application/README.md
```diff
@@ -0,0 +1,28 @@
...
+Invoke with: 
+```bash
+mvn install -Pdeploy-check
+```
+
+## Supported packaging types
+The test application can be packaged using variable packaging types. Select among them using the
+`packaging.type` maven property. These can be mixed with the various profiles above. Supported packaging types:
+- `mvn install -Dpackaging.type=jar`
+- `mvn install -Dpackaging.type=war`
```
meltsufin: Don't you need a different POM file to build the war correctly?

cassand:
I would suggest to rename the artifact to test-spring-application.

I suspect that Spring may interfere with some of the functionalities that we are testing such as the [sessions](https://github.com/spring-projects/spring-session) and maybe the logging, therefore, we might want to later develop an alternative test application without frameworks.

java-runtimes-common/test-application/README.md
```diff
@@ -0,0 +1,28 @@
...
+## Supported profiles
+- **Integration test profile** - stages the application for use in the [runtimes-common integration testing framework](https://github.com/GoogleCloudPlatform/runtimes-common/tree/master/integration_tests).
+Invoke with: 
+```bash
+mvn install -Pint-test
+```
+- **Deploy check profile** - stages the application for deployment directly to App Engine.
+Invoke with: 
+```bash
+mvn install -Pdeploy-check
```
balopat: I'd call this 'deployment-test'

java-runtimes-common/test-application/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+                  </resources>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
+      <!-- this profile stages the application for deploy-check tests -->
```
balopat: "for deployment tests"

java-runtimes-common/test-application/README.md
aslo: Nope, the spring-boot-maven-plugin handles that automatically, depending on the module's `<packaging>` element, which just reads from the `packaging.type` property.

java-runtimes-common/test-spring-application/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
+      <!-- this profile stages the application for deployment tests -->
+      <id>deployment-test</id>
+      <properties>
+        <app.yaml.file>app.deploy-check.yaml</app.yaml.file>
```
meltsufin: Are these two profiles (deployment-test and int-test) even warranted, when the only difference is the `app.yaml.file` property? Also, I still find it confusing in terms of naming because both of these profiles support integration tests. It's just that one is local and another is remote, or am I missing something?

aslo:
@balopat @meltsufin , I've revisited this and made some changes - mind taking another look?

Now, there's a bit less duplication, and the various maven profile names are more descriptive (instead of `-Pint-test` and `-Pdeploy-check`, it's now `-Pruntime.custom` and `-Pruntime.java`. Thanks!

java-runtimes-common/test-spring-application/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+This is a test application to be used in testing the Java Runtimes containers. It implements a set
+of http endpoints which exercise integrations with various Stackdriver APIs, in line with the 
+[Runtimes-common integration test framework](https://github.com/GoogleCloudPlatform/runtimes-common/tree/master/integration_tests#tests).
+
+The test application is a simple [Spring Boot](https://projects.spring.io/spring-boot/) application,
+which can be built in several different ways.
+
+## Supported profiles
+**Custom runtime profile** - stages the application for deployment to a custom runtime on App Engine Flexible:
```
meltsufin: Can you please clarify what "stages" means, like that it prepares files in target/deploy ready to be used by "gcloud app deploy"?

meltsufin: Yes, please.

PR #129, opened 2017-08-10
Implement OpenJDK 9 Image

aslo:
This PR adds an openjdk:9 image, which can be built alongside the existing openjdk:8 image. 

Other notable changes: 
- Created two new maven modules, `openjdk-common` and `openjdk-test-common` to store common resources and tests between the two images.
- Modified the `build.sh` script to require a module (openjdk8 or openjdk9)to be specified, as we will be building/releasing only one image at a time
- Moved `structure_test.sh` and `local_integration_test.sh` into the integration-test phase of the maven build. This simplifies both the maven and the `cloudbuild.yaml` configuration

Still TODO:
 - In order to run `local_integration_test.sh`, docker client must be installed on the builder image:
    - [x] bug fix in javac/docker base image https://github.com/GoogleCloudPlatform/cloud-builders/pull/108
    - [x] set javac/docker image as the maven builder's base image https://github.com/GoogleCloudPlatform/cloud-builders/pull/111


cloudbuild.yaml
```diff
@@ -2,19 +2,15 @@
...
-- name: 'gcr.io/gcp-runtimes/structure_test'
-  args: ['--image', '${_IMAGE}', '-v', '--config', '/workspace/openjdk8/target/test-classes/structure.yaml']
-  id: 'STRUCTURE_TEST'
+  args:
+  - 'clean'
+  - 'install'
+  - '--batch-mode'
+  - '-Ddocker.image.name=${_IMAGE}'
+  # only build the specified module
+  - '--projects=${_MODULE}'
```
aslo: It would appear not - the mvn executable is able to deal with arguments of the form "foo=bar".

scripts/local_integration_test.sh
```diff
@@ -34,26 +34,29 @@ if [[ -z "$imageUnderTest" ]]; then
...
 # build app container locally
 pushd $deployDir
-export STAGING_IMAGE=$imageUnderTest
-envsubst < Dockerfile.in > Dockerfile
+# escape special characters in docker image name
+STAGING_IMAGE=$(echo $imageUnderTest | sed -e 's/\//\\\//g')
+sed -e "s/FROM .*/FROM $STAGING_IMAGE/" Dockerfile.in > Dockerfile
 echo "Building app container..."
-docker build -t $APP_IMAGE . || gcloud docker -- build -t $APP_IMAGE .
+docker build -t $APP_IMAGE . || docker -- build -t $APP_IMAGE .
```
aslo: :+1:

FYI this means we'd need to use a maven/docker cloudbuild step, like we do in jetty, but that shouldn't be too difficult to set up.

meltsufin: I guess one option is to add `gcloud` only if the Cloud SDK installation is detected.

openjdk9/src/test/resources/openjdk9-structure.yaml
```diff
@@ -0,0 +1,13 @@
+# Test spec for the Cloud Runtimes container structure testing framework.
+#
+# See https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/structure_tests/README.md for more detail.
+#
+schemaVersion: 1.0.0
+
+commandTests:
+- name: 'correct java version is installed'
+  command: ['java', '-version']
+  expectedError: ['openjdk version "${openjdk.version.major}-Debian"']
```
balopat: This is interesting. They changed the -version output: https://blogs.oracle.com/java-platform-group/a-new-jdk-9-version-string-scheme and http://openjdk.java.net/jeps/223 -- we should follow this with our parsing logic so that it stays maintainable and we don't lose out on ensuring the minor version. 

Based on above I'm confused by the ``-Debian`` part - where did that get introduced. It also might be part of the major version?

```
openjdk version "9-Debian"
OpenJDK Runtime Environment (build 9-Debian+0-9b181-2)
OpenJDK 64-Bit Server VM (build 9-Debian+0-9b181-2, mixed mode)

```
java 8:
````
Start command: java -showversion -Xms51137M -Xmx51137M -XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+PrintCommandLineFlags -version
-XX:InitialHeapSize=53621030912 -XX:MaxHeapSize=53621030912 -XX:+ParallelRefProcEnabled -XX:+PrintCommandLineFlags -XX:+UseG1GC 
openjdk version "1.8.0_131"
OpenJDK Runtime Environment (build 1.8.0_131-8u131-b11-1~bpo8+1-b11)
OpenJDK 64-Bit Server VM (build 25.131-b11, mixed mode)
````

Comment (aslo) [inline openjdk9/src/test/resources/openjdk9-structure.yaml]: The `-Debian` suffix likely is there due to the fact that we're installing openjdk currently from the "unstable" debian package repository. I'm not sure if debian builds of openjdk9 will always have this suffix, but I'd hope at least that they will follow the new version scheme spec. I haven't seen clear info on this yet.

File: openjdk9/src/main/docker/Dockerfile [hunk 4]
```diff
@@ -0,0 +1,57 @@
...
+FROM ${docker.appengine.image}
+ENV DEBIAN_FRONTEND noninteractive
+
+# Create env vars to identify image
+ENV OPENJDK_VERSION ${openjdk.version}
+ENV GAE_IMAGE_NAME ${project.artifactId}
+ENV GAE_IMAGE_LABEL ${docker.tag.long}
+
+RUN \
+ # add debian 'sid' repo in order to install openjdk-9
```
balopat: typo? this is not the 'sid' anymore it's buster. Maybe a link https://wiki.debian.org/DebianBuster would be good too.

openjdk9/src/main/docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+ # add debian jessie, jessie-backports repo in order to install downgraded libcups2 dependency
+ && echo 'deb http://httpredir.debian.org/debian jessie main' > /etc/apt/sources.list.d/jessie.list \
+ && echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/apt/sources.list.d/jessie-backports.list \
+
+ && apt-get -q update \
+ && apt-get -y -q --no-install-recommends install \
+    # jdk and its dependencies
+    ca-certificates-java=20170531'*' \
+    openjdk-9-jdk-headless=${openjdk.version}'*' \
+    libcups2=2.2.1-8~bpo8+1 \
```
balopat: can we put in a comment above or here, why we need to downgrade libcups2? 
I would even double check if we can remove it completely! (unless of course we want to explicitly support printing from cloud deployed apps or rather the JDK has some weird dependency on CUPS)

openjdk9/src/main/docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+    # system utilities
+    procps \
```
balopat: why do we need procps? jdk8 doesn't have it!

openjdk9/src/main/docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+    netbase \
+    unzip \
+ && apt-get clean \
+ && rm /var/lib/apt/lists/*_*
+
```
balopat: btw jdk8 has wget, jdk9 doesn't. Why the discrepancy and why is it needed at all in jdk8? can we remove it? (maybe a separate issue)

aslo:
@meltsufin @cassand @balopat I've addressed all remaining TODOs and I believe I've addressed all PR comments. Mind giving this another review?

openjdk9/src/main/docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+# Add the cloud debugger libraries
+ADD https://storage.googleapis.com/cloud-debugger/appengine-java/current/cdbg_java_agent.tar.gz /opt/cdbg/
```
balopat: is this the same for jdk8 and 9? Btw, I can't find this URL anywhere documented, is this specifically built for GAE-java runtimes? The official docs from the debugger point to the GCE link: https://storage.googleapis.com/cloud-debugger/compute-java/debian-wheezy/cdbg_java_agent_gce.tar.gz

openjdk9/src/main/docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+COPY docker-entrypoint.bash /
+COPY setup-env.d /setup-env.d/
+COPY shutdown/ /shutdown/
+RUN tar Cxfvz /opt/cdbg /opt/cdbg/cdbg_java_agent.tar.gz --no-same-owner \
+ && rm /opt/cdbg/cdbg_java_agent.tar.gz \
+ && chmod +x /docker-entrypoint.bash /shutdown/*.bash /setup-env.d/*.bash \
+ && mkdir /var/log/app_engine \
+ && chmod go+rwx /var/log/app_engine
```
balopat: all this is completely duplicated across jdk8 and 9 - I'm not sure we can get rid of this easily but it might worth to have another thought about some kind of templating setup.

openjdk9/src/main/docker/Dockerfile
aslo: :+1: You're right, it's not necessary. Created https://github.com/GoogleCloudPlatform/openjdk-runtime/issues/135 to track removing it from openjdk:8.

openjdk9/src/main/docker/Dockerfile
aslo: I'd think this should be the same, but the fact that that URL references wheezy gives me pause. I'll ask around for the current best practice.

aslo:
:+1: Thanks @balopat ! The `out` directory was committed by mistake - good catch.

openjdk9/src/main/docker/Dockerfile
aslo: Fair point.

It looks like some of the openjdk images on dockerhub actually [generate static dockerfiles](https://github.com/docker-library/openjdk/blob/1dce6353dcbb15f09479e7043f48082051967139/9-jdk/Dockerfile#L1-L5), but then they commit them so it's completely transparent and easy to understand. Maybe that's something we can look into in the future.

balopat: that is a cool approach, I created https://github.com/GoogleCloudPlatform/openjdk-runtime/issues/136 we can continue to discuss it there.

README.md
```diff
@@ -1,4 +1,7 @@
-![Build Status](http://storage.googleapis.com/java-runtimes-kokoro-build-badges/openjdk-runtime-master.png)
+![Build Status](http://storage.googleapis.com/java-runtimes-kokoro-build-badges/openjdk-runtime-master.png) 
+
+![Stability](https://img.shields.io/badge/gcr.io/google--appengine/openjdk:8-stable-green.svg)
+![Stability](https://img.shields.io/badge/gcr.io/google--appengine/openjdk:9-unstable-red.svg)
 
 # Google Cloud Platform OpenJDK Docker Image
```
balopat: [in README.md](https://github.com/GoogleCloudPlatform/openjdk-runtime/pull/129/files#diff-04c6e90faac2675aa89e2176d2eec7d8R10): 
````
 This image is mirrored at both `launcher.gcr.io/google/openjdk8` and `gcr.io/google-appengine/openjdk`.
````
I think we'll need to update this for openjdk9 as well I guess? How do we differentiate between 8 and 9 in the google-appengine version?

Comment (balopat) [inline README.md]: Also, how does 8 vs 9 affect ``runtime:java``?

Comment (aslo) [inline README.md]: Yep, we should update this. I'm not sure what's involved in getting images added to `launcher.gcr.io`. @meltsufin Is this something we should look into?

Comment (aslo) [inline README.md]: `runtime: java` is still correct. The user selects between openjdk8/9 using `runtime_config.jdk` (see below). Until we release the openjdk:9 image and add it to the runtime builder, `jdk: openjdk9` won't be available. I'll update the readme to reflect the current state.

```yaml
runtime_config:
  jdk: openjdk8
```

balopat: As discussed, the structure test is a valid positioning for this - however, this case, please rename the test-app to java9-compilation-test or something similar.

balopat:
if we add new JVM options which are specific for the openjdk-9 runtime (see for example https://github.com/GoogleCloudPlatform/openjdk-runtime/issues/137) - where would they fit in this multi-module design?

aslo:
I'd think they should live in the src/main/docker dir within the module that they're specific to. Then the dockerfile would have access to them during the docker build. If we wanted to set additional JVM flags, we could add a script to the setup-env.d directory that modifies the args accordingly.

balopat:
Great, all in my opinion, this is all solid work, great job! LGTM

openjdk9/src/main/docker/Dockerfile
aslo: Just wanted to circle back on this - this is indeed the correct path for the app engine cloud debugger binary. There are no public docs because it's installed automatically and transparently to the user.

aslo:
@meltsufin Do you mind giving this another review so it can be merged? Thanks!

PR #130, opened 2017-08-11
fixes integration test flakiness

### LINKED ISSUE #125: Allow passing app deploy version in App Engine integration test
See: `gcloud app deploy` line in `ae_integration_test.sh`.
Passing a static app version can enable faster deploy cycle.

### LINKED ISSUE #127: Fix flaky monitoring integration test
The monitoring integration test is flaky. 
The following are the main reasons: 

1. The driver can't find the custom metric

BadRequest: 400 The provided filter doesn't refer to any known metric. (GET https://monitoring.googleapis.com/v3/projects/java-runtime-test/timeSeries/?filter=metric.type+%3D+%22custom.googleapis.com%2FUzLlwfHpNOCsZSuu%22&interval.endTime=2017-08-03T21%3A13%3A00.000000Z&interval.startTime=2017-08-03T21%3A11%3A00.000000Z)

Solution: reproduce the error and identify a workaround (work out timings, retries, double check if the metric is stored, etc.)

2. The gRPC call fails while storing the metric

exit code: 500, text: {"timestamp":1501793413295,"status":500,"error":"Internal Server Error","exception":"com.google.api.gax.grpc.ApiException","message":"io.grpc.StatusRuntimeException: INTERNAL: HTTP/2 error code: INTERNAL_ERROR\nReceived Rst Stream","path":"/monitoring"}

Solution: file a bug with the Stackdriver team and hunt down the root cause, and in the mean time catch the exception and retry (maybe double check if the metric was stored)

balopat:
**Improves integration tests** (closes #127)
- retries the custom metric creation when it fails (3 times max)
- pinned temporarily to the latest version of runtimes-common/integration tests published to the java-runtime-tes repo 
- improved logging 
- introduced app checking for GAE to avoid 502 errors on the test itself
- with these settings I have 10/10 success rate with both GKE and GAE integration tests finally

**Enables app version to be passed in for GAE** (closes #125)
- introduced the version deployment with version specific url and ``-no-promote`` to potentially speed up local experimenting (I haven't seen a crazy significant amount of speedup, as the bulk of the deployment time is still spent updating the container I guess, but the separation seems still nice and definitely skips the global traffic rerouting step)

### DIFF
--- a/scripts/ae_integration_test.sh
+++ b/scripts/ae_integration_test.sh
@@ -25,8 +25,18 @@ deployDir=$testAppDir/target/deploy
 
 imageUnderTest=$1
 if [ -z "${imageUnderTest}" ]; then
-  echo "Usage: ${0} <image_under_test>"
+  echo "Usage: ${0} <image_under_test> [gae_deployment_version]"
   exit 1
+fi
+
+# for local tests it makes sense sometimes to pin the deployment to an
+# active version as that will speed up the deployment, for CI/CD this feature
+# is not recommended
+gaeDeploymentVersion=$2
+if [ "${gaeDeploymentVersion}" ]; then
+    DEPLOYMENT_OPTS="-v $gaeDeploymentVersion --no-promote"
+    DEPLOYMENT_VERSION_URL_PREFIX="$gaeDeploymentVersion-dot-"
+
 fi
 
 # build the test app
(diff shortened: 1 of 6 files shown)


test-application/src/main/java/com/google/cloud/runtimes/stackdriver/StackDriverMonitoringService.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.logging.Logger;
+
+@Service
+public class StackDriverMonitoringService {
+
+    private static Logger LOG = Logger.getLogger(StackDriverMonitoringService.class.getName());
+
+    public void createMetricAndInsertTestToken(String projectId, String metricType, long metricValue) throws IOException {
+        MetricServiceClient metricServiceClient = MetricServiceClient.create();
+        int retries = 3;
```
meltsufin: I know it's a test app, but maybe make this constant a property in `application.properties`?

scripts/ae_integration_test.sh
```diff
@@ -25,10 +25,20 @@ deployDir=$testAppDir/target/deploy
...
-  echo "Usage: ${0} <image_under_test>"
+  echo "Usage: ${0} <image_under_test> [gae_deployment_version]"
   exit 1
 fi
 
+# for local tests it makes sense sometimes to pin the deployment to an
+# active version as that will speed up the deployment, for CI/CD this feature
+# is not recommended
+gaeDeploymentVersion=$2
+if [ ! -z "${gaeDeploymentVersion}" ]; then
```
aslo: I believe `[ -n $foo ]` or just `[ $foo ]` are equivalent to `[ ! -z $foo ]`, and might be a bit cleaner.

test-application/src/main/java/com/google/cloud/runtimes/stackdriver/StackDriverMonitoringService.java
balopat: sounds good, maybe a little bit overkill for a test-app, but definitely forward looking :)
fixed.

scripts/ae_integration_test.sh
balopat: thanks, good spot. fixed.

PR #133, opened 2017-08-14
adding deployment token as a check before integration tests

balopat:
Mitigates further race conditions that result in 502 on AppEngine and hung tests for GKE.


aslo:
Ah, this is clever. Nice!

PR #134, opened 2017-08-17
adding custom logging test + local runtimes-common integration test

### LINKED ISSUE #126: Setup local integration test with the runtimes-common integration testing
Via running the ``gcr.io/gcp-runtimes/integration_test`` image locally (same functionality as in integration_test.yaml), we could actually iterate much faster when implementing changes on the test-app. 
Currently the local_integration_test.sh is only testing whether the app starts and verifies the shutdown logs.

### LINKED ISSUE #140: Use local container builder for running integration tests locally
https://github.com/GoogleCloudPlatform/container-builder-local

### LINKED ISSUE #145: Structure tests fails (maven property missing) with Maven 3.0.5
During  `mvn clean install` the following error show up
```
status: Image is up to date for gcr.io/gcp-runtimes/structure_test:latest
2017/09/01 15:00:07 Running tests for file /cfg/openjdk8-structure.yaml
--- FAIL: TestAll (0.21s)
    --- FAIL: TestAll/Command_Test:_correct_java_version_is_installed (0.12s)
        structure_test_v1.go:171: Executing: [java -version]
        structure_test_v1.go:188: stderr: openjdk version "1.8.0_131"
                OpenJDK Runtime Environment (build 1.8.0_131-8u131-b11-1~bpo8+1-b11)
                OpenJDK 64-Bit Server VM (build 25.131-b11, mixed mode)
        structure_test.go:56: Expected string 'openjdk version "1\.${openjdk.version.major}.*${openjdk.version.minor}"' not found in error!
    --- FAIL: TestAll/Command_Test:_correct_javac_version_is_installed (0.09s)
        structure_test_v1.go:171: Executing: [javac -version]
        structure_test_v1.go:188: stderr: javac 1.8.0_131
        structure_test.go:56: Expected string 'javac 1\.${openjdk.version.major}.*${openjdk.version.minor}' not found in error!
        structure_test.go:45: Total tests run: 2

```

The generated `target/test-classes/openjdk8-structure.yaml` have the following content:
``` yaml
schemaVersion: 1.0.0

commandTests:
- name: 'correct java version is installed'
  command: ['java', '-version']
  expectedError: ['openjdk version "1\.${openjdk.version.major}.*${openjdk.version.minor}"']
- name: 'correct javac version is installed'
  command: ['javac', '-version']
  expectedError: ['javac 1\.${openjdk.version.major}.*${openjdk.version.minor}']
```

I think that the maven properties `openjdk.version.major` and `openjdk.version.minor` are not correctly set or interpreted by maven.

The issue is also present for the `openjdk9` project

balopat:
* This the first half of #5 enabling [custom logging tests](https://github.com/GoogleCloudPlatform/runtimes-common/tree/master/integration_tests/#custom-logging). The standard logging tests are going to be in the next PR. 
   * Important to note: https://github.com/GoogleCloudPlatform/runtimes-common/pull/308 is a prerequisite for this, otherwise the driver is not able to pick up the project id from the docker container (as it's using gcloud for that which is currently not setup in the docker container with the right information). However, for now I published my branch version to the java-runtimes-test docker registry that can serve as our version of the driver until we negotiate the PR with @nkubala.
  * closes #126 
  * closes #145 
  * closes #140

TODO left: 

- [x] fix local runtimes-common integration tests with local cloud builder (if possible)
- [x] fix logging test timeout issue (recently came up) 
- [x] final review and refactor (opened new ticket for a larger redesign: #148)


scripts/local_integration_test.sh
```diff
@@ -44,48 +90,66 @@ pushd $deployDir
 export STAGING_IMAGE=$imageUnderTest
 envsubst < Dockerfile.in > Dockerfile
 echo "Building app container..."
-docker build -t $APP_IMAGE . || gcloud docker -- build -t $APP_IMAGE .
+docker build -t $APP_IMAGE . || gcloud docker -- build -t $APP_IMAGE . || usage "Error building test-app image from base image \"${imageUnderTest}\", please make sure it exists!"
+
+
+if [[ "$WITH_GCP" ]]; then
```
aslo: I don't think it makes sense to me to add these integration tests to `the local_integration_test.sh` script. I think we originally added this script in order to test things that were outside the scope of the runtimes_common integraion testing framework (namely, container shutdown hooks).

I think if we wanted to make it easier to run the runtimes_common tests, a better way to do it would be to just simulate a cloudbuild execution locally, using the `container-builder-local` emulator that was just released as part of the Cloud SDK (see [release notes](https://cloud.google.com/sdk/docs/release-notes#google_cloud_container_builder)).

test-application/src/main/java/com/google/cloud/runtimes/config/MockGCPConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.IOException;
+import java.util.logging.Logger;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+
+@Configuration
+@Profile("mock-gcp")
```
aslo: I'm skeptical about having these separate profiles for gcp and mock-gcp. Is it possible to write a single test app implementation, and have the environment in which the test app runs handle the responsibility of passing in google credentials?

balopat: One of the reasons I introduced this was that wihtout mock-gcp - the ``local_integration_test.sh`` (and I mean the shutdown log testing too, not just the local runtime-commons test) would need to pass in the GCP creds, otherwise the app doesn't start up. Is it okay to expect a valid GCP setup for the user in the local_integration_test.sh?

scripts/local_integration_test.sh
balopat: oooo, I'll test out this ``container-build-local``, thanks for the suggestion! 
I created a separate script first btw, but then I figured that they were so similar that I integrated it back into the local_integration_test.sh. It's really handy to have as part of the workflow. But if container-build-local can do the magic, I'll have a look at that as that could remove the duplication around the arguments of the runtimes-common driver for example (one place here, the other place is the yaml file).

test-application/src/main/java/com/google/cloud/runtimes/config/MockGCPConfiguration.java
aslo: Maybe the `local_integration_test.sh` script could check for the presence of the `GOOGLE_CREDENTIALS` environment variable, and either fail or warn the user if it looks like the environment wasn't properly set up. What do you think about that?

balopat: It's doable, the only thing I have against that is that this decision will force someone to setup a service account even when they start their development...which might not be a too strong issue but we'll have to document it then.

balopat: @meltsufin what do you think?

test-application/src/main/java/com/google/cloud/runtimes/LoggingTestController.java
```diff
@@ -0,0 +1,84 @@
+package com.google.cloud.runtimes;
+
+import com.google.cloud.MonitoredResource;
+import com.google.cloud.logging.LogEntry;
+import com.google.cloud.logging.Logging;
+import com.google.cloud.logging.Payload;
+import com.google.cloud.logging.Severity;
+import com.google.cloud.runtimes.stackdriver.StackDriverMonitoringService;
```
meltsufin: Unused import

test-application/src/main/java/com/google/cloud/runtimes/LoggingTestController.java
```diff
@@ -0,0 +1,84 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+import sun.util.logging.resources.logging;
```
meltsufin: Unusued import

test-application/src/main/java/com/google/cloud/runtimes/LoggingTestController.java
```diff
@@ -0,0 +1,84 @@
...
+            return "LoggingTestRequest{" +
+                    "level='" + level + '\'' +
+                    ", log_name='" + log_name + '\'' +
+                    ", token='" + token + '\'' +
+                    '}';
+        }
+    }
+
+    @RequestMapping(path = "/logging_custom", method = POST)
+    public String handleMonitoringRequest(@RequestBody LoggingTestRequest loggingTestRequest) throws IOException, InterruptedException {
```
meltsufin: `handleLoggingRequest`?

test-application/src/main/java/com/google/cloud/runtimes/LoggingTestController.java
```diff
@@ -0,0 +1,84 @@
...
+        List<LogEntry> entries = new ArrayList<>();
+        Payload.StringPayload payload = Payload.StringPayload.of(loggingTestRequest.token);
+        Severity severity = Severity.valueOf(loggingTestRequest.level);
+        LogEntry entry = LogEntry.newBuilder(payload)
+                .setSeverity(severity)
+                .setLogName(loggingTestRequest.log_name)
+                .setResource(MonitoredResource.newBuilder("global").build())
+                .build();
+        entries.add(entry);
+        logging.write(entries);
```
meltsufin: Instead of using the Logging client lib directly, we should be able to use the JUL appender. Can you also look at #72 that has been put on the back-burner, and reconcile it with what you're doing in this PR? Ideally, we would like to close out #72 as well, but I think there are some unresolved complications there with setting up JUL.

test-application/src/main/java/com/google/cloud/runtimes/config/GCPConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+
+import java.io.IOException;
+
+import static com.google.cloud.ServiceOptions.getDefaultProjectId;
+
+@Configuration
+@Profile("gcp")
+public class GCPConfiguration {
```
meltsufin: `GcpConfiguration`

test-application/src/main/java/com/google/cloud/runtimes/config/MockGCPConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+public class MockGCPConfiguration {
```
meltsufin: `MockGcpConfiguration`

test-application/src/main/java/com/google/cloud/runtimes/config/MockGCPConfiguration.java
meltsufin: Can we just not invoke this logging test when running locally? Another option would be to have a separate test-app that doesn't depend on GCP for local testing. Otherwise, I don't really mind the mock-gcp profile.

test-application/src/main/java/com/google/cloud/runtimes/LoggingTestController.java
cassand: I think that the Custom Logging Test should use the client library and make an explicit call to Stackdriver Logging (by creating only one LogEntry), whereas the Standard Logging Test should redirect the output of JUL to Stackdriver.

test-application/src/main/java/com/google/cloud/runtimes/config/MockGCPConfiguration.java
balopat: The problem is that I am using Spring to inject the services, so when the beans get created, they need the credentials and thus _the application fails to start_ without the creds - that's why I went down the road to create the two profiles. I just had an idea that I'll check out: I should try to make the components ``@Lazy``.
Re: skipping the GCP related tests locally: if we combine @aslo's idea of separating the runtimes-common tests out into a separate script, then this will naturally skip calling them in the local integration tests.
So, if I can make ``@Lazy`` work, we have a "single test app implementation" (without need for profiles), the app starts up and we'll have the runtimes-common tests separated, which might need the GCP creds.

test-application/src/main/java/com/google/cloud/runtimes/LoggingTestController.java
balopat: I agree with @cassand - I was planning to use JUL for standard logging tests and Logging library directly for custom, just so that we have both covered. 
However, we are free to decide what we want to do, as we can setup custom logging with JUL too.

scripts/local_integration_test.sh
balopat: I will do this in #126 though, and then separate it out from this pull request.

test-application/src/main/java/com/google/cloud/runtimes/LoggingTestController.java
meltsufin: I'm OK with Damien's suggestion.

balopat:
@aslo @meltsufin can you please review again this PR? I addressed all of your concerns: 
- @aslo: removed the separate MockGCP profile and introduced container-builder-local for local testing
- @meltsufin: cleanup of unused imports and method rename (handleLoggingRequest) is done 

Thanks!

scripts/local_runtimes_common_integration_test.sh
```diff
@@ -0,0 +1,111 @@
...
+OUTPUT_FILE=${CONTAINER}-output.txt
+DEPLOYMENT_TOKEN=$(uuidgen)
+
+readonly imageUnderTest=$1
+if [[ -z "$imageUnderTest" ]]; then
+  echo "Usage: ${0} <image_under_test>"
+  exit 1
+fi
+
+if [[ -z ${GOOGLE_APPLICATION_CREDENTIALS} ]]; then
```
meltsufin: The readme instructions don't mention `GOOGLE_APPLICATION_CREDENTIALS`.
Can we just leverage gcloud for the default credentials?

scripts/local_runtimes_common_integration_test.sh
```diff
@@ -0,0 +1,111 @@
...
+# build app container locally
+pushd $deployDir
+export STAGING_IMAGE=$imageUnderTest
+envsubst < Dockerfile.in > Dockerfile
+echo "Building app container..."
+docker build -t $APP_IMAGE . || gcloud docker -- build -t $APP_IMAGE .
+
+docker rm -f $CONTAINER || echo "Integration-test-app container is not running, ready to start a new instance."
+
+# run app container locally to test shutdown logging
```
meltsufin: You're not really doing a shutdown logging test here.

test-application/pom.xml
```diff
@@ -35,13 +35,33 @@
...
       <version>0.7.0</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.api</groupId>
+      <artifactId>gax</artifactId>
+      <version>1.6.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.api</groupId>
+      <artifactId>gax-grpc</artifactId>
```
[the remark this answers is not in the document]
balopat: good catch - let me double check & I will clarify in comments.

PR #143, opened 2017-08-31
AE integration tests use unique version IDs

aslo:
* App engine deployments now use unique, random version strings. This makes it effectively impossible for concurrent integration test runs to collide. 
* Added a cleanup cloudbuild execution to tear down resources that are spun up for testing.

### DIFF
--- a/scripts/ae_integration_test.sh
+++ b/scripts/ae_integration_test.sh
@@ -30,15 +30,14 @@ if [ -z "${imageUnderTest}" ]; then
   exit 1
 fi
 
-# for local tests it makes sense sometimes to pin the deployment to an
-# active version as that will speed up the deployment, for CI/CD this feature
-# is not recommended
-readonly gaeDeploymentVersion=$2
-if [ "${gaeDeploymentVersion}" ]; then
-    DEPLOYMENT_OPTS="-v $gaeDeploymentVersion --no-promote"
-    DEPLOYMENT_VERSION_URL_PREFIX="$gaeDeploymentVersion-dot-"
-
+# If provided, pin to a specific version. Otherwise, generate a random new version ID to prevent
+# inadvertent collisions.
+gaeDeploymentVersion=$2
+if [ -z "${gaeDeploymentVersion}" ]; then
+  gaeDeploymentVersion=$(uuidgen)
 fi
+DEPLOYMENT_OPTS="-v $gaeDeploymentVersion --no-promote"
+DEPLOYMENT_VERSION_URL_PREFIX="$gaeDeploymentVersion-dot-"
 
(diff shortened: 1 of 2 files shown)


scripts/ae_integration_test.sh
```diff
@@ -77,3 +76,10 @@ gcloud container builds submit \
   --substitutions "_DEPLOYED_APP_URL=$DEPLOYED_APP_URL" \
   $dir
 
+# run a cleanup build once tests have finished executing
```
meltsufin: Can we do the cleanup only if the version was autogenerated?
Otherwise, we'll lose the benefit of fast deployments to the same running version.

scripts/ae_integration_test.sh
```diff
@@ -30,15 +30,14 @@ if [ -z "${imageUnderTest}" ]; then
...
-    DEPLOYMENT_OPTS="-v $gaeDeploymentVersion --no-promote"
-    DEPLOYMENT_VERSION_URL_PREFIX="$gaeDeploymentVersion-dot-"
-
+# If provided, pin to a specific version. Otherwise, generate a random new version ID to prevent
+# inadvertent collisions.
+gaeDeploymentVersion=$2
+if [ -z "${gaeDeploymentVersion}" ]; then
+  gaeDeploymentVersion=$(uuidgen)
 fi
+DEPLOYMENT_OPTS="-v $gaeDeploymentVersion --no-promote"
```
meltsufin: Do we also need to make sure that the previous version is not stopped as well, or is that the default behavior anyway?

scripts/ae_integration_test.sh
aslo: Ah yep, good call!

scripts/ae_integration_test.sh
aslo: I don't think it would make a difference if the version was stopped or not. What is your concern exactly?

meltsufin: If the tests are run in parallel, they may stop each other's versions.

scripts/ae_integration_test.sh
meltsufin: You can also just add an argument to specify whether to kill the version.

scripts/ae_integration_test.sh
aslo: Oh, right. I'll add the --no-stop-previous-version flag just to be sure.

scripts/ae_integration_test.sh
aslo: I'm a little wary of adding more flags - I'd prefer to kill the versions only if we're also creating them in this script.

PR #146, opened 2017-09-01
Fix random string generation

aslo:
`uuidgen` will not work as a means for generating valid GKE cluster names - they must start with a letter and be shorter than 40 chars. Instead, we should just use a string of 20 random lowercase letters.


cassand:
Wouldn't a timestamp (prefixed by a letter) be preferable ?

aslo:
The issue with timestamps is that two of these builds will often be launched at roughly the same time, so it's likely that there will be identical timestamps for each. Doing it this way means it's very unlikely to have collisions.

PR #153, opened 2017-09-13
Fix war deployments for test app

aslo:
Makes a few more final tweaks that are required in order to package the test app as a WAR file that works on jetty/tomcat webservers. Because some dependency management is required to produce a Spring Boot WAR file, the JAR/WAR switch was moved from a maven property to a maven profile.

See also https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#howto-create-a-deployable-war-file

### DIFF
--- a/java-runtimes-common/test-spring-application/pom.xml
+++ b/java-runtimes-common/test-spring-application/pom.xml
@@ -26,7 +26,6 @@
     <deployment.token>default_deployment_token</deployment.token>
     <packaging.type>jar</packaging.type>
     <app.deploy.runtime>java</app.deploy.runtime>
-    <app.deploy.image></app.deploy.image>
   </properties>
 
   <dependencies>
--- a/java-runtimes-common/test-spring-application/pom.xml
+++ b/java-runtimes-common/test-spring-application/pom.xml
@@ -24,6 +24,9 @@
     <maven.compiler.source>${java.version}</maven.compiler.source>
     <maven.compiler.target>${java.version}</maven.compiler.target>
     <deployment.token>default_deployment_token</deployment.token>
+    <packaging.type>jar</packaging.type>
+    <app.deploy.runtime>java</app.deploy.runtime>
+    <app.deploy.image></app.deploy.image>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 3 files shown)


java-runtimes-common/test-spring-application/pom.xml
```diff
@@ -25,6 +25,8 @@
     <maven.compiler.target>${java.version}</maven.compiler.target>
     <deployment.token>default_deployment_token</deployment.token>
     <packaging.type>jar</packaging.type>
+    <app.deploy.runtime>java</app.deploy.runtime>
+    <app.deploy.image></app.deploy.image>
```
balopat: why do we need an empty property to be defined?

aslo: 👍  Can be omitted. FYI if the property is missing, the template variable will remain in the generated Dockerfile, but come to think of it, this is actually better than what I have here, where it will just be filled in with an empty string.

java-runtimes-common/test-spring-application/src/main/java/com/google/cloud/runtimes/Application.java
```diff
@@ -2,10 +2,18 @@
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.builder.SpringApplicationBuilder;
+import org.springframework.boot.web.support.SpringBootServletInitializer;
 
 @SpringBootApplication
-public class Application {
+public class Application extends SpringBootServletInitializer {
```
balopat: `SpringBootServletInitializer` is required for the WAR, I understand that.  Just to make sure, did you test that the executable JAR is still working with the embedded server?

aslo: I did run integration tests with the new test app, and it appears to work the same as before in JAR form. From the [spring boot javadocs](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/web/support/SpringBootServletInitializer.html), the language they use seems to imply that it doesn't break anything to add this, and my testing seems to indicate the same.

PR #154, opened 2017-09-20
Run integration tests as part of image release build

aslo:
This change executes integration tests from within the Cloud Container Builder execution that builds releases. This ensures that no release candidates will be pushed without first having passed integration tests.

This is analogous to the work done for the jetty runtime in https://github.com/GoogleCloudPlatform/jetty-runtime/pull/189.


balopat:
I would argue that we should keep "master" always green, thus putting the pressure of integration tests onto the "merge to mainline" process instead of the "release process" - hence release becomes (stays) relatively painless. Good CI/CD practices would point to this direction. If you have a trustworthy line of commits in master, you can release anytime.

aslo:
@balopat That's a good point. If you think this adds unnecessary overhead to the release process I'm happy to abandon this change.

balopat:
I think the criteria of "having release candidates passing integration tests" is a must. However I would rather implement it with better Github merge checks that ensure integration tests are passing before merging to `master`. This creates a set of "trusted commits". 
However, as I'm thinking about it - these "trusted commits" might get out of date between the time they were merged to master and the release, due to changes in the infrastructure. If we see the chances for this divergence to be significant - then it makes sense to derisk it with an "release time" integration test.  
Otherwise it's an overhead. 
@meltsufin what do you think?

meltsufin:
Things do change between commits and the release. So, I think it's a good idea to run integration tests right before release. However, I also hear the concern about it slowing down the release process, which is already quiet painful. If the tests only take a few minutes to run, I'm fine with this.

scripts/build.sh
```diff
@@ -22,6 +22,7 @@ usage() {
   echo "             -d|--docker-namespace <docker_namespace> - a docker repository beginning with gcr.io"
   echo "             -m|--module           <module_to_build>  - one of {openjdk8, openjdk9}"
   echo "           [ -s|--tag-suffix ]                        - suffix for the tag that is applied to the built image"
+  echo "           [ -p|--project ]                           - GCP test project to use for staging images. If not provided, the default GCP project will be used."
```
meltsufin: If the user can provide the project, can we get rid of the `--docker-namespace` because it would just be `gcr.io/${project}`?

balopat:
Alright, let's go with the conservative approach and put it in - if we're worried that something changes, the wait should be worthwile even if it takes 10+ minutes.

aslo:
SGTM. FYI, these tests will add a minimum of 10 minutes to the release process.

scripts/build.sh
balopat: local build is failing for a while now, it hasn't been introduced here, it was introduced with local-shutdown-test being part of the maven build :( 
See #147 to resolve it strategically, or remove the shutdown-test from the maven build.

meltsufin: Ok, I forgot the issue was still not resolve.d

aslo: The `--docker-namespace` arg refers to the final built image, while `--project` is a project to use for staging only. I think we need both.

java-runtimes-common/mvn-gcloud-image/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+        --bash-completion=false \
+        --disable-installation-options && \
+
+    /builder/google-cloud-sdk/bin/gcloud config set component_manager/disable_update_check 1 && \
+
+    # Kubernetes configuration
+    /builder/google-cloud-sdk/bin/gcloud config set compute/zone us-east1-b && \
+    /builder/google-cloud-sdk/bin/gcloud components install kubectl -q
+
+ENV PATH=/builder/google-cloud-sdk/bin/:$PATH
```
balopat: can't we just publish this image upfront and depend on it as a contract rather than having it rebuilt on every build?

cloudbuild.yaml
```diff
@@ -13,4 +14,24 @@ steps:
...
+# Tag and push the openjdk image to a staging repository for integration testing
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['tag', '${_IMAGE}', '${_STAGING_IMAGE}']
+  id: 'TAG_STAGING_IMG'
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['push', '${_STAGING_IMAGE}']
+  id: 'PUSH_STAGING_IMG'
+
+# Build a 'builder' image for running integration tests
+- name: 'gcr.io/cloud-builders/docker'
```
balopat: yeah, I'm getting convinced that it is not necessary to run this step every time we want to build the image. 
What do you think?

aslo: Agreed. Adding issue #155 to track this, and I'll add a followup PR right after this one to use the prebuilt image.

aslo:
@meltsufin Added documentation updates.

scripts/build.sh
meltsufin: Do we ever use a different project for staging vs publishing?
Even if that's the case, I think it would make sense to just have `--staging-project` and `--publishing-project`, just for the sake of symmetry.

aslo: :+1: added --publishing-project and --staging-project.

meltsufin:
@aslo Can you also update the `DEVELOPING.md` with the argument changes? Thanks!

aslo:
@meltsufin Done!

aslo:
@meltsufin I think this should be good to merge! PTAL when you have a chance.

PR #159, opened 2017-10-09
unpinning ca-certificates-java on buster

balopat:


balopat:
Thanks, good pointer. I'll create a ticket for that - openjdk8 should be on stretch and the new ca-certs-java version then. For [openjdk9](https://github.com/docker-library/openjdk/blob/8a23a228bda7d2edeb4132fffd2d08c1e1fcf4ac/9-jdk/Dockerfile#L56) it seems there is no need for pinning

PR #168, opened 2017-12-19
Jdk9 cacerts fix

ChengyuanZhao:


openjdk9/src/main/docker/Dockerfile
```diff
@@ -24,6 +24,10 @@ RUN \
  # add debian stretch-backports repo in order to install openjdk-9
  echo 'deb http://httpredir.debian.org/debian stretch-backports main' > /etc/apt/sources.list.d/stretch-backports.list \
 
+ # openjdk9 for some reason comes without these cacerts, and to successfully reinstall them
+ # they need to be removed here before the subsequent install
+ && dpkg --purge --force-depends ca-certificates-java \
```
meltsufin: Why would `ca-certificates-java` would be installed at this point, before we install any JDK?
Also, which issue is this addressing?

meltsufin: I think the cloud debugger not working is separate?

ChengyuanZhao: This was an issue from our original internal investigation. The `GcpHubClient.java` would call `RegisterDebuggee` method and would get an exception: 

`HubClient.registerDebuggee: javax.net.ssl.SSLException: java.lang.RuntimeException: Unexpected error: 
 java.security.InvalidAlgorithmParameterException: the trustAnchors parameter must be non-empty at`

The stackdriver team found it was caused by this http://www.oracle.com/technetwork/java/javase/9-0-1-relnotes-3883752.html#JDK-8189131. Also, simply installing ca-certs would not work, and after some googling it turns out people were using the workaround that I have in this PR.

ChengyuanZhao: https://stackoverflow.com/questions/33439905/is-the-cacerts-file-missing-in-ubuntu-15-10-and-openjdk-8-jdk

ChengyuanZhao:
CI failed due to error getting a needed image from gcr.io, closing and reopening to redo the CI build.

ChengyuanZhao:
Hmm actually it works for me without this line too. It looks like Les's problem could be caused by something else.

PR #172, opened 2018-02-23
Add Cloud Profiler agent support

meltsufin:


openjdk-test-common/src/test/resources/structure.yaml
```diff
@@ -67,6 +77,10 @@ fileExistenceTests:
   path: '/opt/cdbg/format-env-appengine-vm.sh'
   isDirectory: false
   shouldExist: true
+- name: 'Cloud profiler agent is installed'
```
aalexand: Stackdriver profiler

openjdk-common/src/main/docker/setup-env.d/25-profiler-env.bash
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# configure Cloud Profiler
+
+export PROFILER_AGENT_COMMAND=${PROFILER_AGENT_COMMAND:-"-agentpath:/opt/cprof/profiler_java_agent.so=--logtostderr"}
+
+export PROFILER_AGENT=
+
+if is_true "$PROFILER_ENABLE"; then
+  PROFILER_AGENT=${PROFILER_AGENT_COMMAND}
```
aalexand: We should skip setting this if $JAVA_USER_OPTS contains profiler_java_agent.so string to avoid double-initializing the agent for Alpha users where the instructions to enable the profiler were use JAVA_USER_OPTS.

meltsufin:
Thanks for the review @aalexand!

openjdk-test-common/src/test/resources/workspace/setup-profiler.bash
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+source /setup-env.d/05-utils.bash
+
+function testProfilerOn() {
+  source /setup-env.d/25-profiler-env.bash
+  if [ "$PROFILER_AGENT" != "-agentpath:/opt/cprof/profiler_java_agent.so=--logtostderr" ]; then
```
balopat: this is a duplication of the string literal above - when we decide to change it a little bit around (e.g. increase threadcount, change logging, etc.), it will be very easy to miss this setup and can cause issues later. What do you think about refactoring the command piece into a separate file, like `profile_agent` and then you could do `"$PROFILER_AGENT" != "$(cat profile_agent)"`...?

openjdk-test-common/src/test/resources/workspace/setup-java.bash
```diff
@@ -60,10 +60,11 @@ JAVA_TMP_OPTS=-XX:Temp
 JAVA_HEAP_OPTS=-XX:Heap
 JAVA_GC_OPTS=-XX:GC
 DBG_AGENT=debug
+PROFILER_AGENT=profiler
 JAVA_USER_OPTS=user
 
 source /setup-env.d/30-java-env.bash
-if [ "$(echo $JAVA_OPTS | xargs)" != "-showversion -XX:Temp debug -XX:Heap -XX:GC user" ]; then
+if [ "$(echo $JAVA_OPTS | xargs)" != "-showversion -XX:Temp debug profiler -XX:Heap -XX:GC user" ]; then
```
balopat: I know this is there for a while now, but why do we need the xargs?

openjdk-test-common/src/test/resources/workspace/setup-java.bash
```diff
@@ -102,7 +103,7 @@ if [ "$TEST_MIN_HEAP" != "OK" -o "$TEST_MAX_HEAP" != "OK" ]; then
 fi
 
 
-#test GKE envrionment
+#test GKE environment
```
balopat: there are a bunch of places where this file unsets all the parameters, that should be extracted to a function + we should include the `PROFILER_AGENT` variable in it as well.

openjdk-test-common/src/test/resources/workspace/setup-profiler.bash
meltsufin: This file is actually used for tests only, and it's duplicated on purpose for testing.

openjdk-test-common/src/test/resources/workspace/setup-java.bash
meltsufin: I think it was there just to allow easier comparison. Again, this is just used in tests. I would ignore this issue as part of this PR.

balopat:
LGTM

PR #176, opened 2018-05-02
Add back libgnutls30.

### LINKED ISSUE #166: Refactor structure tests to use intermediate container for test scripts
See discussion here: https://github.com/GoogleCloudPlatform/container-structure-test/issues/9

### LINKED ISSUE #177: Cannot invoke wget due to libgnutls missing
Repro:
```console
$ docker run --entrypoint bash --rm -it gcr.io/google-appengine/openjdk
root@ce985d89b5bc:/# wget google.com
wget: error while loading shared libraries: libgnutls.so.30: cannot open shared object file: No such file or directory
```

dlorenc:
Fixes #177.
Fixes #166.

### DIFF
--- a/openjdk9/src/main/docker/Dockerfile
+++ b/openjdk9/src/main/docker/Dockerfile
@@ -34,10 +34,7 @@ RUN \
     # other system utilities
     netbase \
     unzip \
-
- # Remove libgnutls30 dependency brought in by openjdk package. One of its dependencies uses
- # the AGPL license, so we cannot use it.
- && dpkg --remove --force-depends libgnutls30 \
+    wget \
 
  # cleanup package manager caches
  && apt-get clean \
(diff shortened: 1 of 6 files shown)


openjdk9/src/main/docker/Dockerfile
```diff
@@ -35,10 +35,6 @@ RUN \
     netbase \
     unzip \
 
- # Remove libgnutls30 dependency brought in by openjdk package. One of its dependencies uses
- # the AGPL license, so we cannot use it.
- && dpkg --remove --force-depends libgnutls30 \
```
meltsufin: While, we're here, we should probably also add `wget` above to be consistent with `openjdk:8`.

balopat:
Ok, this is going to be a much longer exercise than I thought. Upgrading to the new structure tests is tricky as we heavily depended on external scripts as checkers: https://github.com/GoogleCloudPlatform/openjdk-runtime/issues/166
This is the list of scripts we'll need to now provide as image: 
https://github.com/GoogleCloudPlatform/openjdk-runtime/tree/master/openjdk-test-common/src/test/resources/workspace

balopat:
hmm, it wasn't so bad after all...I created a temporary image in our wrapper script adding the "workspace" as a folder - it worked locally, let's see if Travis passes or not.

dlorenc:
Woohoo!

PR #195, opened 2019-04-23
Remove openjdk:11 mention until we are actually delivering it

donmccasland:


wuservices:
I `openjdk11` still not supported? I saw it mentioned in the README, but I'm wondering it's not intended to be there yet:

> To select a jdk version, use the `runtime_config.jdk` field in app.yaml. Supported JDK versions include `openjdk8` and `openjdk11`.

easeev:
Any ETA for JDK 11 support?

PR #196, opened 2019-04-30
Added MAX_HEAP_SIZE_MB to avoid a calculated size being too large

gregw:
When deployed on systems with large available memory, the automatic calculated heap size may be too large for the JVM.
This PR applies a limit to the calculated heap size.


meltsufin:
@gregw Why is setting `HEAP_SIZE_MB` directly or adjusting the ratio not sufficient?

gregw:
@meltsufin I think the image should work out of the box on all reasonable containers.   Currently I can't even build on a machine with 32GB because the tests fail because they try to start a JVM with a too large heap.
If we are setting the heap size by heuristics, then those heuristics should always result in a legal sensible heap size.  Overriding values should only be needed if you want a very specific configuration.

PR #209, opened 2022-09-07
Switching to read openjdk8 from adoptium

pmaloogoogle:


jinglundong:
That travis-ci is broken after they moved to another domain. https://b.corp.google.com/issues/244516861

openjdk8/src/main/docker/Dockerfile
```diff
@@ -25,18 +25,35 @@ RUN \
...
     # install the jdk and its dependencies
     ca-certificates-java \
-    openjdk-${openjdk.version.major}-jdk-headless=${openjdk.version}'*' \
     # procps is used in the jvm shutdown hook
     procps \
     # other system utilities
     netbase \
     wget \
     unzip \
+    fontconfig \
```
jinglundong, line 34: Do you remember which command requires this? We want to keep this list of package as small as possible. If required, maybe add a comment?

openjdk8/src/main/docker/Dockerfile
```diff
@@ -25,18 +25,35 @@ RUN \
...
 
+ARG OPENJDK_VER='jdk8u345-b01'
+ARG OPENJDK_VER_NAME='8u345b01'
+ARG OPENJDK_TEMP_PATH='/tmp/openjdk'
+ARG OPENJDK_OUT='/usr/lib/openjdk'
+ENV PATH $OPENJDK_OUT/bin:$PATH
+ENV JAVA_HOME $OPENJDK_OUT
+
+RUN mkdir -p $OPENJDK_TEMP_PATH \
+ && mkdir -p $OPENJDK_OUT \
```
jinglundong, line 49: Any preferences on && over another `RUN` here? With one `RUN` for each command, the error message will more explicit when it fails. There might be performance impact, but I wouldn't imagine it's going to be significant enough.

jinglundong (approved):

openjdk8/src/main/docker/Dockerfile
pmaloogoogle: I picked && to follow the existing patterns we use in the Dockerfile. I also remember Ludo mentioned that single RUN creates single layer.

openjdk8/src/main/docker/Dockerfile
ludoch: Would keep it, was there in the adoptium dockerfile.

pmaloogoogle: I added fontconfig and locales per Ludo's recommendation. Don't think it is required right now. @ludoch  - thoughts on removing this?

openjdk8/src/main/docker/Dockerfile
ludoch: combining RUN is better for minimizing layers.

ludoch (approved):

